PR #74, opened 2021-05-03
Documentation for CHKV

albertmink:


file-formats/chkv/format.md
```diff
@@ -0,0 +1,9 @@
+# File Format for Object Type CHKV
+
+The file representation for object `CI_SEARCH_ABAP_PATTERN` is provided [here](./examples/ci_search_pattern.chkv.json).
```
schneidermic0, line 3: Shouldn't we stick to a Z-example like `z_aff_example_chkv`

schneidermic0 (changes requested):
I would prefer Z-objects as examples

schneidermic0 (approved):
Thanks, looks good to me.

PR #81, opened 2021-05-05
Change implementingClass to an local example class

Christopher-Hermann:
Usage of a productive check class should be avoided since we added a 1:1 relation between CHKO and CLAS


bulcu (approved):

albertmink (approved):
Thank you.

PR #89, opened 2021-05-12
Specification of file format for ATC check categories

wurzka:


albertmink (approved):
Thank you! Look's good to me.

PR #129, opened 2021-07-20
Update ENHS Type

### LINKED ISSUE #117: No ABAP type for ENHS, yet
The schema for object type ENHS is not based on an ABAP type, yet.

wurzka:
closes #117


file-formats/enhs/enhs.json
```diff
@@ -36,15 +38,15 @@
           ],
           "enumDescriptions": [
             "Standard",
```
marcushoepfner, line 40: Is "Standard" correct or is it "Default"? Not sure. Have seen default as well in the past.

file-formats/enhs/enhs.json
```diff
@@ -132,8 +134,8 @@
             "description": "Example classes of the BAdI definition",
             "type": "array",
             "items": {
-              "title": "Example Class",
-              "description": "Example class",
+              "title": "Object Name",
+              "description": "Object name with max. length 30",
```
marcushoepfner, line 138: this information is duplicated with he maxLength attribute. can you re-phrase the description. mabye just "Object name" is also ok?!

marcushoepfner (commented):
in general there are also other so-called enhancement technologies than badi_def.
that's why there is a tool_type mentioned in the enhancement spot. but I guess we can add more later to the schema (I think we don't need others yet) and have a new "type" next to badi_definitions in ty_main then.

file-formats/enhs/enhs.json
wurzka: @marcushoepfner  do you prefer "Object name" or the old title and description ("Example Class")? For the first option, I have to adopt the interface zif_aff_types_v1 with the reusable components. This would not be a big change, but I'd like to check the other option first.

file-formats/enhs/type/zif_seef_aff_enhs_v1.intf.abap
```diff
@@ -0,0 +1,223 @@
...
+    "! Object type ENHS
+    BEGIN OF ty_main,
+      "! <p class="shorttext">Schema</p>
+      "! Format version
+      "! $required
+      schema           TYPE string,
+      "! <p class="shorttext">Header</p>
+      "! Header
+      "! $required
+      header           TYPE zif_aff_types_v1=>ty_header_60,
```
schneidermic0, line 213: Enhancement Editor allows to enter 100 characters as description. I guess we have to introduce a new header :(

file-formats/enhs/type/zif_seef_aff_enhs_v1.intf.abap
```diff
@@ -0,0 +1,223 @@
...
+      use_fallback_class TYPE abap_bool,
+      "! <p class="shorttext">Name of the Default/Fallback Class</p>
+      "! Name of the default/fallback class
+      fallback_class     TYPE zif_aff_types_v1=>ty_object_name_30,
+      "! <p class="shorttext">Limited Filter Use</p>
+      "! Limited filter use
+      filter_limitation  TYPE abap_bool,
+      "! <p class="shorttext">Documentation Id</p>
+      "! Documentation Id
+      documentation_id   TYPE c LENGTH 60,
```
schneidermic0, line 181: Shouldn't this be 40 characters?

file-formats/enhs/enhs.json
wurzka: Do you mean that as ENHS specific or general? We used "Standard" also for all the other types.

file-formats/enhs/type/zif_seef_aff_enhs_v1.intf.abap
```diff
@@ -0,0 +1,223 @@
...
+      "! <p class="shorttext">Interface of the BAdI Definition</p>
+      "! Interface of the BAdI definition
+      "! $required
+      interface          TYPE zif_aff_types_v1=>ty_object_name_30,
+      "! <p class="shorttext">Instantiation of the BAdI Definition</p>
+      "! Instantiation of the BAdI definition
+      instantiation      TYPE ty_instantiation,
+      "! <p class="shorttext">Single- or Multiple Use BAdI</p>
+      "! Indicator whether the BAdI definition is single- or multiple usable
+      multiple_use       TYPE abap_bool,
```
schneidermic0, line 163: If we touch it now, maybe you want to check discussion in https://github.com/SAP/abap-file-formats/issues/61#issuecomment-839606098

file-formats/enhs/type/zif_seef_aff_enhs_v1.intf.abap
```diff
@@ -0,0 +1,223 @@
...
+    END OF co_filter_type.
+
+  "! <p class="shorttext">Type of the Value Check Object</p>
+  "! Type of the value check object
+  "! $values {@link zif_seef_aff_enhs_v1.data:co_value_check_object_type}
+  TYPES ty_value_check_object_type TYPE string.
+
+  CONSTANTS:
+    "! Type of the value check object
+    "! $values {@link zif_seef_aff_enhs_v1.data:co_value_check_object_type}
```
schneidermic0, line 71: We don't need the annotation `$values` here, do we?

file-formats/enhs/enhs.json
marcushoepfner: for me it is ok to take the one with less effort

file-formats/enhs/enhs.json
marcushoepfner: then it´s fine for me

schneidermic0 (changes requested):
See my comments.

As soon as we have adapted the format I suggest to reintroduce an example which was deleted (currently) by this PR

file-formats/enhs/type/zif_seef_aff_enhs_v1.intf.abap
wurzka: I added a new header zif_aff_types_v1=>ty_header_100

file-formats/enhs/type/zif_seef_aff_enhs_v1.intf.abap
wurzka: The types that are actually used internally have 60 characters

file-formats/enhs/type/zif_seef_aff_enhs_v1.intf.abap
wurzka: That's why a developer should avoid copy/paste...

marcushoepfner (approved):

file-formats/enhs/type/zif_seef_aff_enhs_v1.intf.abap
schneidermic0: OK, Thanks

file-formats/enhs/type/zif_seef_aff_enhs_v1.intf.abap
wurzka: If we changed this to an enum (or rather our "constant enum type"), wouldn't it result in changing all booleans to make it consistent? Or is this a special case?

schneidermic0: Sorry, I missed to add a point from #61. It was about whether empty values shall be displayed. I took this example that we might want to show always whether a BAdI definition is configured as singleUsed or multipleUsed. 
I see three options here:

1. Only show the value, if it is a multipleUsed BAdI
2. We use `$showAlways` here
3. We change it to an enum

For me, all options work. BTW- adding or removing `$showAlways` can be considered as compatible change [here](https://github.com/SAP/abap-file-formats/blob/main/doc/specification.md#compatible-file-format-changes)

schneidermic0 (changes requested):
Looks good to me, except for the removed example. Do you plan to add id?

Just see my additional comment. You can decide whether, you want to change it.

file-formats/enhs/type/zif_seef_aff_enhs_v1.intf.abap
wurzka: Personally, I prefer the $showAlways tag. That's why we introduced it

wurzka:
The example will be added as soon as I get the 'ok' for the type for ENHS.

schneidermic0 (approved):
Here you go :)

file-formats/enhs/type/zif_seef_aff_enhs_v1.intf.abap
```diff
@@ -173,9 +173,6 @@ INTERFACE zif_seef_aff_enhs_v1
       "! <p class="shorttext">Example Classes of the BAdI Definition</p>
       "! Example classes of the BAdI definition
       example_classes    TYPE ty_example_classes,
-      "! <p class="shorttext">Use Fallback Class</p>
-      "! Use fallback class in case no implementation is executed
-      use_fallback_class TYPE abap_bool,
       "! <p class="shorttext">Name of the Default/Fallback Class</p>
       "! Name of the default/fallback class
```
schneidermic0, line 177: Should we add the meaning of the fallback class in the description?
```suggestion
      "! Name of the default/fallback class. The fallback class is executed if no BAdI implementation exists.
```

schneidermic0 (commented):
See my comment

schneidermic0 (approved):
Thanks. looks good to me

PR #136, opened 2021-07-22
CLAS: Correct example

wurzka:
The field `category` is not tagged with `$required` or `$showAlways`. Hence, category `generalObjectType` (ABAP Type 00) shall not be written in serialization process.


crasovan (approved):
ok

PR #156, opened 2021-08-12
Specification, see commit msg

albertmink:
Note: lines with multiple sentences are bad for code-review, L118

### DIFF
--- a/doc/specification.md
+++ b/doc/specification.md
@@ -23,7 +23,7 @@ with the mandatory elements object_name, object_type and file_extension.
 Namespaces are converted into brackets, for instance `/NMSPC/CL_OBJECT` becomes `(nmspc)cl_object`.
 
 ### Object Name and Type
-The object name and type represent the object. Basically, the object name (R3TR) and type are stored in the ABAP object directory (aka TADIR).
+The object name (R3TR) and object type are derived from the object itself, as stored in the ABAP object directory (aka TADIR).
 
 ### Sub Object Name and Sub Object Type
 
--- a/doc/specification.md
+++ b/doc/specification.md
@@ -27,7 +27,7 @@ The object name (R3TR) and object type are derived from the object itself, as st
 
 ### Sub Object Name and Sub Object Type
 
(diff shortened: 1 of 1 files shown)


doc/specification.md
```diff
@@ -23,16 +23,16 @@ with the mandatory elements object_name, object_type and file_extension.
 Namespaces are converted into brackets, for instance `/NMSPC/CL_OBJECT` becomes `(nmspc)cl_object`.
 
 ### Object Name and Type
-The object name and type represent the ABAP object. Basically, the object name (R3TR) and type are stored in the ABAP object directory (aka TADIR).
+The object name (R3TR) and object type are derived from the object itself, as stored in the ABAP object directory (aka TADIR).
 
 ### Sub Object Name and Sub Object Type
 
-Some object types support sub objects (LIMU object types) like function modules in a function group. For sub objects, the main object name and type must be part of the file name.
+Some objects support sub objects (LIMU object) like function modules in a function group. For sub objects, the main object name and type must be part of the file name.
```
schneidermic0, line 30: I am not sure whether we should stick to `object types` here.

doc/specification.md
```diff
@@ -80,7 +80,7 @@ The JSON files additionally follow
 
 ## Format Compatibility
 
-The format of object types will evolve over time. If a change to the format is considered incompatible, the version of the file format has to be increased.
+The format of objects will evolve over time. If a change to the format is considered incompatible, the version of the file format has to be increased.
```
schneidermic0, line 83: I am not sure whether we should stick to `object type` here.

schneidermic0 (commented):
Thanks, Albert. Removing ABAP in front of objects is fine. I am not sure whether we should stick to the term "object type" if we are talking about types and not objects.

doc/specification.md
albertmink: Reverted to the previous version.

doc/specification.md
albertmink: Reverted to the previous version.

schneidermic0 (approved):
Thanks. LGTM :)

PR #170, opened 2021-08-13
Fix filename pattern in specification

schneidermic0:
The filename pattern in the specification was not correct: `<content_type>` and `<language>` are independent of each other.


wurzka (approved):
LGTM :)

PR #174, opened 2021-08-17
remove global in filenames for CLAS

### LINKED ISSUE #30: INTF vs CLAS abap, common file naming?

albertmink:
closes #30


wurzka (approved):
LGTM :)

PR #204, opened 2021-09-01
Fix schema for ENHO

wurzka:
Here was an error that affected the indentation. I have regenerated the schema.


albertmink (approved):
Only whitespace changes.
Look's good to me.

PR #227, opened 2021-09-10
GH action from 80sec to 20 sec

albertmink:
as long as we use from the super-linter suite only its
editorconfig feature, we can do editorconfig-checker directly

actually, super-linter action takes about 80 sec. Now it should be
faster


larshp:
might need someone with access to settings to change the required checks?

image

albertmink:
well, we might need an @SAP/abap-file-formats-admin to  adjust the GitHub actions setting -- see comment above

.github/workflows/editorconfig.yml
```diff
@@ -0,0 +1,35 @@
...
+# https://help.github.com/en/articles/workflow-syntax-for-github-actions
+#
+
+#############################
+# Start the job on all push #
+#############################
+on:
+  push:
+    branches-ignore: [main]
+    # Remove the line above to run when pushing to master
```
larshp:
```suggestion
```

albertmink:
Add new action with this PR.
Delete the old one in a second PR.

larshp (approved):
looks good to me 👍

PR #232, opened 2021-09-17
rm old super-linter action for editorconfig

albertmink:


larshp (approved):

larshp:
guess this also needs changes to required checks

albertmink:
My intuition is telling me the same. However, the configuration says, all actions are allowed 😕

larshp:
allowed, yes

issues is these two where the required part must be removed in branch protection(repo settings)
image

albertmink:
Ok, I got it down to one 
image

What happens if we remove the `required` from the other linter? Would GitHub allow to merge, if linting action fails?

PR #235, opened 2021-09-23
Adds documentation and ABAP types for JSON versioning

albertmink:
Needs updates as we implement the type writer that translates ABAP types into JSON.

### DIFF
--- a/doc/specification.md
+++ b/doc/specification.md
@@ -84,11 +84,20 @@ To enable interchangeability, the files are characterized by
 The JSON files additionally follow
 * 2 spaces indentation
 
-## Format Compatibility
+## Format Versions and Compatibility
 
-The format of object types will evolve over time. If a change to the format is considered incompatible, the version of the file format has to be increased.
+The ABAP file format of object types will evolve over time.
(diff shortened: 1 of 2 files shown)


albertmink:
I struggled to do a proper merge, so please take special care to the diff.

docs/json.md
```diff
@@ -176,10 +174,10 @@ This leads to the following generated JSON schema:
   "description": "Interface properties",
   "type": "object",
   "properties": {
-    "$schema": {
-      "title": "Schema",
+    "formatVersion": {
```
wurzka, line 177: The field schema has to be removed from the ABAP type `ty_main` which is defined above

wurzka (approved):
LGTM. But do we want to merge this already? Only files for interface have been changed. Other still use $schema

albertmink:
Yes, there is only the INTF that has the `new` JSON format. Sooner or later the others will follow.
As we already made the decision and put a document that explains the design, for me this PR is an approximation to this decision.
I tend to merge it and consider it an incremental step.

PR #236, opened 2021-09-23
How JSON versioning is defined (here: INTF)

albertmink:
New JSON Schema and data for the object type INTF, that shows how we defined JSON versioning in ABAP file formats.


schneidermic0 (approved):
If we don't want to add a `$` in front of `formatVersion`, it looks good to me.

wurzka (approved):
LGTM :)

PR #253, opened 2021-10-18
rm additionalProperties from documentation, with update on compatible changes

albertmink:


docs/specification.md
```diff
@@ -87,6 +87,8 @@ Following changes to file formats are considered as compatible:
 
 - Values for a field (enum) are added (remark: this might lead to syntax errors in ABAP systems which don't support this value)
 - Descriptions or titles are changed
+- Non-mandatory fields are added
+- Non-mandatory fields are removed
```
schneidermic0, line 91: I am just wondering wether we shall move the information about compatible and incompatible changes to file docs/json.md.

The statements are related to JSON format changes. But we can change this also in a separate PR.

What do you think?

Issue #248, opened 2021-10-07
JSON schema: Change property `additionalProperties` to `true`


schneidermic0:
The property `additionalProperties` in our JSON schema is set to `false`, currently. Do to this setting it is incompatible to add or remove non-mandatory fields. 

As mentioned in #105, adding and removing non-mandatory fields would be compatible if the property `additionalProperties` is set to `true`. With this we won't have to increase the format version if we add or remove new non-mandatory fields. 

What do you think?

See also: 

- https://github.com/SAP/abap-file-formats/blob/main/docs/specification.md#incompatible-file-format-changes
- #105

albertmink:
We will stay with the restrictive `"additionalProperteries" = false` setting. If we introduce/remove a new ABAP file format version that e.g. adds a field (mandatory or non-mandatory), we include/remove it to the new JSON Schema. This don't break our recent compatibility strategy https://github.com/SAP/abap-file-formats/blob/main/docs/json.md#compatible-file-format-changes

albertmink:
need to partially rewind/revert introduces changes https://github.com/SAP/abap-file-formats/pull/253

schneidermic0:
Today, we had a discussion with the team. We tend to keep the property `additionalProperties` with value `false`. With this we can avoid that random fields are added to the data.
On the other hand, this means we have to consider the removal of fields in general as an incompatible change. But I think this would be fine. 

Any objections?

If there are no objections, I'll prepare a change to update the documentation accordingly.

schneidermic0:
We decided to keep the field `additionalProperties`. Therefore, I close this issue.

PR #254, opened 2021-10-18
rm `additionalProperties` in JSON Schema

albertmink:
still needs the adoption in our dev systems (type_writer)


albertmink:
simple manipulation with `sed` was not enough, need to rework

albertmink:
see #248

PR #268, opened 2021-11-30
Change type of bufferedNumbers to integer

wurzka:
Now, the type of bufferedNumbers is integer with minimum 0 and maximum 99999999. This corresponds to former type n length 8.


schneidermic0 (approved):
Reviewed the change also with the domain owners.

LGTM

PR #269, opened 2021-12-01
Delete part with default title and description

wurzka:
We decided **not** to write a default title or a default description if no title or description is provided. This had to be updated in our documentation.


schneidermic0 (approved):
Thanks. LGTM

PR #272, opened 2021-12-13
Update compatibility rules

wurzka:


wurzka:
I was very unsure how to formulate this so that it is understandable what we want to say.

docs/json.md
```diff
@@ -25,14 +25,17 @@ If a change to the format is considered incompatible, then the `formatVersion` h
```
albertmink, line 25:
```suggestion
For compatible changes to the file format, the format is just updated.
The indicator whether the format has to be changed should not be updated.
```

albertmink (approved):
Content wise, it think it is an improvement. I noted that we say something about the compatibility in two places. Maybe it is worth to clean it up (here or in an other PR). I see two options for where we can say the compatibility thing:
* Separately in the section `Compatible File Format Changes` and `Incompatible File Format Changes`.
* Put the compatibility thing in the above section.

wurzka:
I would prefer having it separately in the two sections. Nevertheless, it doesn't disturb me that it was already mentioned before that formatVersion is increased in case of an incompatible change.

schneidermic0 (approved):
LGTM

PR #293, opened 2022-01-31
Move sentence to top

huber-nicolas:
I think that this sentence should be at the top. Previously it was directly after the $default annotation


wurzka:
I decided to put it after the descriptions of the different annotations. Hence, there was first the definition and then the rules how to use it

PR #296, opened 2022-02-03
JSON validation: If empty, don't write it

albertmink:


albertmink:
This commit removes the line at the bottom if there is nothing to show.
image

PR #315, opened 2022-03-04
Update JSON specification

### LINKED ISSUE #300: Enum and missing defaults
Importing a JSON data file with enums without specified default value in the JSON Schema may result in an undefined behaviour. E.g., if the JSON data file doesn't contain a specific value for the enum, it is not defined which one to choose.  

See #298

wurzka:
Closes #300

I added a remark that enums should not be used without initial value, required or default


albertmink (approved):
🙏

PR #327, opened 2022-03-16
Update example chkc

huber-nicolas:
In the system we have added a parent category to the example chkc


albertmink (approved):
I'm not sure about the name of the parent CHKC. If there are problems, then we simply rename it.

PR #338, opened 2022-03-21
[RONT] New Object Type SAP Object Type

Ehret-Thomas:
Added new repo folder for object type RONT to AFF repository.


file-formats/ront/ront-v1.json
```diff
@@ -0,0 +1,130 @@
...
+          "maxLength": 2,
+          "pattern": "^[a-z]+$"
+        },
+        "abapLanguageVersion": {
+          "title": "ABAP Language Version",
+          "description": "ABAP language version",
+          "type": "string",
+          "enum": [
+            "standard",
+            "keyUser",
```
schneidermic0, line 39: Does RONT support ABAP language version `keyUser`?

file-formats/ront/type/zif_aff_ront_v1.intf.abap
```diff
@@ -0,0 +1,64 @@
...
+      "! <p class="shorttext">Header</p>
+      "! Header
+      "! $required
+      header         TYPE zif_aff_types_v1=>ty_header_60,
+
+      "! <p class="shorttext">Category</p>
+      "! Category
+      "! $required
+      "! $values {@link zif_aff_ront_v1.data:co_category}
+      type_category  TYPE c LENGTH 2,
```
schneidermic0, line 57: Is it `type_category` or just `category`? The title and description might indicate that you want to name it just `category`?

file-formats/ront/type/zif_aff_ront_v1.intf.abap
```diff
@@ -0,0 +1,64 @@
...
+
+      "! <p class="shorttext">SAP Object Node Type</p>
+      "! $required
+      rontnodes      TYPE SORTED TABLE OF ty_rontnode WITH NON-UNIQUE KEY node_name,
```
schneidermic0, line 61: `rontnodes` sounds very technical to me. Is this the intended name? What is the relation to the title "SAP Object Node Type"? For a list of entries/array/standard table I would expect the title as plural.

schneidermic0 (commented):
@Ehret-Thomas Thanks for your contribution. 

I have added some questions below.

file-formats/ront/README.md
```diff
@@ -0,0 +1,5 @@
+# RONT File Format
+
+File | Cardinality | Definition | Schema | Example
+:--- | :---  | :--- | :--- | :---
+`<name>.ront.json` | 1 | [`zif_aff_ront_v1.intf.abap`](./type/zif_aff_ront_v1.intf.abap) | [`ront-v1.json`](./ront-v1.json) | [`zxxxxxx.ront.json`](./examples/zxxxxxx.ront.json)
```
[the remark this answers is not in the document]
Ehret-Thomas: Hi Michael,
first of all, thanks a lot for your good questions and suggestions. We are currently debugging an issue with our EML-based implementation of the persistence API for the SAP Object Type (RONT). This is why there is no sample object of type RONT yet that I could upload to the repo. I will soon add the required example object. Within this week, I will provide a SQL-based implementation of the relevant persistence methods, following the example of object type Event Binding (EVTB). If we can't resolve the issue in our EML-coding, I will switch to the SQL-coding to be able to generate first RONT objects. I'm sorry for the inconvenience.

file-formats/ront/ront-v1.json
Ehret-Thomas: I asked Martin and my other colleagues from RAP Services about this. For now RONT does not support ABAP Language Version "keyUser".

file-formats/ront/type/zif_aff_ront_v1.intf.abap
Ehret-Thomas: You are right about that inconsistency. I will switch "category" consistently to "type category".

file-formats/ront/type/zif_aff_ront_v1.intf.abap
Ehret-Thomas: To start with, a short explanation of the admittedly strange"RONT". Naming the four-letter transport type "RONT" was due to the fact that "SONT" was unfortunately no longer available. Our next best idea was to pick an "R" instead of "S" for RAP, and  the "N" for node was necessary to get to four letters.   

I will change "SAP Object Node Type" to "SAP Object Node Types". Thanks!

"rontnodes" stems from the name of the corresponding persistence table with the "s" added to make clear that there can be multiple nodes. The persistence tables (including the table RONT_NODE), in turn, were named analogously to the persistence tables of object type "EVTB".

file-formats/ront/type/zif_aff_ront_v1.intf.abap
```diff
@@ -0,0 +1,64 @@
...
+
+      "! <p class="shorttext">Name</p>
+      "! Raw (camel case) node name
+      "! $required
+      node_name_raw TYPE c LENGTH 40,
+
+      "! <p class="shorttext">Description</p>
+      "! Description
+      "! $required
+      description   TYPE c LENGTH 60,
```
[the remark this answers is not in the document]
Ehret-Thomas: Good point. I will make this change and the other ones from above, regenerate the repo folder, and push it to the repository.

file-formats/ront/README.md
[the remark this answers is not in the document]
schneidermic0: Cool. Thanks

file-formats/ront/type/zif_aff_ront_v1.intf.abap
```diff
@@ -0,0 +1,64 @@
...
+  TYPES:
+    "! <p class="shorttext">General Information</p>
+    BEGIN OF ty_main,
+      "! $required
+      format_version TYPE zif_aff_types_v1=>ty_format_version,
+
+      "! <p class="shorttext">Header</p>
+      "! Header
+      "! $required
+      header         TYPE zif_aff_types_v1=>ty_header_60,
```
schneidermic0, line 51: This changes the ABAP Language Version to `standard` and `cloud` (without `keyUser`).
```suggestion
      header         TYPE zif_aff_types_v1=>ty_header_60_cloud,
```

file-formats/ront/type/zif_aff_ront_v1.intf.abap
```diff
@@ -0,0 +1,71 @@
...
+  TYPES:
+    "! <p class="shorttext">Node</p>
+    BEGIN OF ty_rontnode,
+      "! <p class="shorttext">Name</p>
+      "! Node name
+      "! $required
+      node_name     TYPE c LENGTH 40,
+
+      "! <p class="shorttext">Root Node Flag</p>
+      "! Root node flag
```
huber-nicolas, line 29: Here we could give a more detailled description

file-formats/ront/type/zif_aff_ront_v1.intf.abap
```diff
@@ -0,0 +1,71 @@
...
+      "! $required
+      is_root       TYPE abap_bool,
+
+      "! <p class="shorttext">Name</p>
+      "! Raw (camel case) node name
+      "! $required
+      node_name_raw TYPE c LENGTH 40,
```
huber-nicolas, line 36: @schneidermic0 should we put a TYPES ty_object_name_40 TYPE c LENGTH 40 to if_aff_types_v1? What is your opinion?

file-formats/ront/type/zif_aff_ront_v1.intf.abap
```diff
@@ -0,0 +1,71 @@
...
+      "! $values {@link zif_aff_ront_v1.data:co_category}
+      type_category  TYPE c LENGTH 2,
+
+      "! <p class="shorttext">Root SAP Object Node Type</p>
+      "! Root SAP object node type
+      "! $required
+      root_node_type TYPE ty_rontnode,
+
+      "! <p class="shorttext">SAP Object Node Types</p>
+      "! $required
```
huber-nicolas, line 67: If you make the table required, then f.ex. in the example RONT an empty array is printed into the json when the node_types arrays would be empty:
"nodeTypes": [ ]

In my opinion it is sufficient to have the fields of structure ty_rontnode on required.

file-formats/ront/type/zif_aff_ront_v1.intf.abap
```diff
@@ -0,0 +1,71 @@
...
+    "! <p class="shorttext">Node</p>
+    BEGIN OF ty_rontnode,
+      "! <p class="shorttext">Name</p>
+      "! Node name
+      "! $required
+      node_name     TYPE c LENGTH 40,
+
+      "! <p class="shorttext">Root Node Flag</p>
+      "! Root node flag
+      "! $required
```
huber-nicolas, line 30: Is this field really required?
If we remove the required tag then "isRoot": false, is not printed into the json only "isRoot": true is printed to the json. As I understood in the node_types table "isRoot" is always false and is also not editable to true, so we can ommit "isRoot" (atleast in the node_types table). I would remove the required tag

huber-nicolas: Do we even need the "isRoot" information or can we remove it?

Is it working like this?

image

file-formats/ront/type/zif_aff_ront_v1.intf.abap
schneidermic0: If we have 3 usages or more, we should move it to the common interface. Otherwise, I would keep it in the domains.

file-formats/ront/type/zif_aff_ront_v1.intf.abap
schneidermic0: I would have the same question. Since we distinguish between root node types and node types, do you still need the field `is_root`?

file-formats/ront/type/zif_aff_ront_v1.intf.abap
schneidermic0: I agree with @huber-nicolas as long empty `nodeTypes` are allowed.

file-formats/ront/examples/z_aff_example_ront.ront.json
```diff
@@ -0,0 +1,34 @@
...
+    "nodeName": "Z_AFF_EXAMPLE_RONT",
+    "isRoot": true,
+    "nodeNameRaw": "rootNodeType",
+    "description": "Short text description for root SAP Object Node Type."
+  },
+  "nodeTypes": [
+    {
+      "nodeName": "NODETYPE1",
+      "isRoot": false,
+      "nodeNameRaw": "nodeType1",
```
schneidermic0, line 18: As far as I understood `nodeName` and `nodeNameRaw` can have completely different content (like in the root node), don't they. Otherwise (if the conenten is the same and differ only in the case) one can discuss whether the `rawNodeName` would be sufficient, because from this we can "calculate" the `nodeName`. Developers would have to specify it only once.

schneidermic0 (commented):
@huber-nicolas asked me to have a look at this PR. 

I think you made good progress/improvements. Thanks. 👍  

See my comments

file-formats/ront/type/zif_aff_ront_v1.intf.abap
Ehret-Thomas: Yes, you are right. By distinguishing between the root SAP Object Node Type in component root_node_type and the other SAP Object Node Types in table nodeTypes the information about which node type "is root" is already encoded. This makes the component is_root unnecessary. 
Thanks for making this good point!

file-formats/ront/examples/z_aff_example_ront.ront.json
Ehret-Thomas: Very good point. After discussing this question, we decided to enforce that the nodeNameRaw converted upper case has to be equal to the node_name. There will be a validation ensuring this.  
Making this requirement, the node_name can now be calculated from node_name_raw, so that it no longer needs to be part of the AFF interface.

file-formats/ront/type/zif_aff_ront_v1.intf.abap
Ehret-Thomas: Thanks for pointing this out. A SAP Object Type with a root SAP Object Node Type only, that is with empty nodeTypes is allowed. I will remove the $required ABAP Doc comment for the table nodeTypes.

huber-nicolas (approved):
Looks good to me. I only have one remark

file-formats/ront/type/zif_aff_ront_v1.intf.abap
```diff
@@ -0,0 +1,60 @@
...
+      configuration_object TYPE c LENGTH 2 VALUE `CO`,
+    END OF co_category.
+
+  TYPES:
+    "! <p class="shorttext">Node</p>
+    BEGIN OF ty_rontnode,
+      "! <p class="shorttext">Name</p>
+      "! Raw (camel case) node name
+      "! $required
+      node_name_raw TYPE c LENGTH 40,
```
schneidermic0, line 26: I like that you reduced it to just one node name. Thanks. 

If this is the external representation, would you also like to keep the `_raw` in the end or does it make sense to call it just `node_name`?

file-formats/ront/type/zif_aff_ront_v1.intf.abap
```diff
@@ -0,0 +1,60 @@
...
+
+      "! <p class="shorttext">Description</p>
+      "! Description
+      "! $required
+      description   TYPE zif_aff_types_v1=>ty_description_60,
+    END OF ty_rontnode.
```
schneidermic0, line 32: Minor: Does it make sense to rename the type to `ty_node_type`?

```suggestion
    BEGIN OF ty_node_type,
      "! <p class="shorttext">Name</p>
      "! Raw (camel case) node name
      "! $required
      node_name_raw TYPE c LENGTH 40,

      "! <p class="shorttext">Description</p>
      "! Description
      "! $required
      description   TYPE zif_aff_types_v1=>ty_description_60,
    END OF ty_node_type.
```

file-formats/ront/type/zif_aff_ront_v1.intf.json
```diff
@@ -0,0 +1,7 @@
+{
+  "formatVersion": "1",
+  "header": {
+    "description": "RAP SOT:  SAP Object Type AFF Interface",
```
schneidermic0, line 4: Minor ;)
```suggestion
    "description": "RAP SOT: SAP Object Type AFF Interface",
```

schneidermic0 (commented):
@Ehret-Thomas Cool improvements. It looks much cleaner now.

Just see my comments and questions. I think all of these are only minor

file-formats/ront/examples/z_aff_example_ront.ront.json
```diff
@@ -0,0 +1,26 @@
+{
+  "formatVersion": "1",
+  "header": {
+    "description": "Example description for SAP Object Type.",
+    "originalLanguage": "en"
+  },
+  "typeCategory": "businessObject",
+  "rootNodeType": {
+    "nodeNameRaw": "z_Aff_Example_Ront",
```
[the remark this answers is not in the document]
Ehret-Thomas: I agree. "Z" looks better. I've made the change.

file-formats/ront/type/zif_aff_ront_v1.intf.abap
Ehret-Thomas: Yes, I think that's a good idea. It's more consistent with the component names root_node_type and node_types and is easier to understand for someone looking at the interface for the first time. Thanks!

file-formats/ront/type/zif_aff_ront_v1.intf.abap
Ehret-Thomas: Yes, that makes sense to me. I would suggest to use `name` instead of `node_name`, however, as this avoids potential confusion between external and internal node names in the persistence coding for object type RONT.

schneidermic0 (approved):
Thanks @Ehret-Thomas. LGTM

huber-nicolas (approved):

PR #247, opened 2021-10-01
Upload srvb latest

ParashuramaV:

PR #339, opened 2022-03-22
[SRVB] New object type service bindings

wurzka:
follow up from #247


file-formats/srvb/srvb-v1.json
```diff
@@ -0,0 +1,153 @@
...
+          "maxLength": 2,
+          "pattern": "^[a-z]+$"
+        },
+        "abapLanguageVersion": {
+          "title": "ABAP Language Version",
+          "description": "ABAP language version",
+          "type": "string",
+          "enum": [
+            "standard",
+            "keyUser",
```
huber-nicolas, line 40: is key user possible?

wurzka: According to the view you gave me, it is possible

huber-nicolas (approved):

PR #352, opened 2022-04-21
Update documentation for overwriting enum values

### LINKED ISSUE #312: Overwriting enum values
Enums in the JSON Schema are generated out of a type referring to a constant. So far, the values of the enum are given by the names of the components of the constant after being transformed to camel case. 

However, the object type DMON brings up the need for a possibility to manually set the enum values:

Screenshot 2022-02-28 at 08 34 40

Here, it would be better, to have enum values `"CLAS"`, `"FUGR"` and `"PROG"`, which is actually not possible.

wurzka:
Now, enum values can be overwritten by using an annotation. I updated the documentation to describe the procedure

closes #312

### DIFF
--- a/docs/json.md
+++ b/docs/json.md
@@ -160,7 +160,7 @@ If not the transformed name of a component shall be used, the enum value can be 
 ```abap
 "! $enumValue 'newEnumValue'
 ```
-to the specific component. 
+to the specific component.
 
 Titles and descriptions of the enum values are passed to the JSON Schema in the same way as described above.
 They are written in the fields `enumTitles` and `enumDescriptions`.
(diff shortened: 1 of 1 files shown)


albertmink (approved):
Looks good to me.

docs/json.md
```diff
@@ -155,6 +155,13 @@ The type specifies the underlying data type and links to the constant via the fo
 ```
 The names of the components of the constant are written as external (JSON) values to the JSON Schema after being transformed to camel case (e.g, component `badi_definition` is transformed to the enum value `badiDefinition` in the JSON Schema).
 The corresponding values of the components represent the internal (ABAP) values.
+
+If not the transformed name of a component shall be used, the enum value can be overwritten by adding an ABAP Doc comment with the annotational keyword
+```abap
+"! $enumValue 'newEnumValue'
+```
+to the specific component.
+
```
schneidermic0, line 164: Does it make sense to mention that we recommend to use this addition rarely?

schneidermic0 (commented):
Thanks. See my comment

docs/json.md
wurzka: I think this is a good idea. See my next commit

schneidermic0 (approved):
LGTM, Thanks.

PR #354, opened 2022-04-22
[LIMU INDX] File format for DDIC secondary indexes

jheinri:
AFF file format for ABAP DDIC secondary indexes


file-formats/tabl/type/zif_aff_indx_v1.intf.abap
```diff
@@ -0,0 +1,42 @@
+INTERFACE zif_aff_indx_v1
+  PUBLIC.
+
+  TYPES:
+    "! <p class="shorttext">Index Fields</p>
+    "! Index fields
+    BEGIN OF ty_indexfield,
+      "! <p class="shorttext">Field Name</p>
+      "! Field name
+      fieldname TYPE c LENGTH 30,
```
schneidermic0, line 10: I guess the field name should be required, shouldn't it?

Minor: I also suggest to rename it to field_name so that we get the key "fieldName"

```suggestion
      "! $required
      field_name TYPE c LENGTH 30,
```

file-formats/tabl/type/zif_aff_indx_v1.intf.abap
```diff
@@ -0,0 +1,42 @@
...
+      "! <p class="shorttext">Field Position</p>
+      "! Position
+      position  TYPE n LENGTH 4,
```
schneidermic0, line 13: Question: Must the field position be specified by the user or is this just calculated or just the position in the table?
If we used the order in the field list, we most probably would not be able to use a `sorted table`

file-formats/tabl/README.md
```diff
@@ -0,0 +1,6 @@
+# TABL File Format
+
+File | Cardinality | Definition | Schema | Example
+:--- | :---  | :--- | :--- | :---
+`<name>.tabl.json` | 1 | *does not exist, yet* | *does not exist, yet* | *does not exist, yet*
+`<name>.tabl.<indexname>.indx.json` | 0..n | [`zif_aff_indx_v1.intf.abap`](./type/zif_aff_indx_v1.intf.abap) | [`indx-v1.json`](./indx-v1.json) | [`zaff_exmpl_tabl.tabl.001.indx.json`](./examples/zaff_exmpl_tabl.tabl.001.indx.json)
```
schneidermic0, line 6: If we don't add an example in the first step, we might have to remove the links.
```suggestion
`<name>.tabl.<indexname>.indx.json` | 0..n | [`zif_aff_indx_v1.intf.abap`](./type/zif_aff_indx_v1.intf.abap) | [`indx-v1.json`](./indx-v1.json) | `zaff_exmpl_tabl.tabl.001.indx.json`
```

schneidermic0 (changes requested):
Thanks @jheinri for your contribution 👍 

Please see my questions and comments

larshp:
and again, why not just make this part of TABL, it is logically part of TABL, the implementation is part of TABL?

schneidermic0:
Independent how we solve it (as part of the *.tabl.json-file or as dedicated file(s)) I would prefer to keep it also as close as possible how it is/will be provided in the development environment.

file-formats/tabl/type/zif_aff_indx_v1.intf.abap
```diff
@@ -0,0 +1,42 @@
...
+      format_version TYPE zif_aff_types_v1=>ty_format_version,
+      "! <p class="shorttext">Header</p>
+      "! Header
+      "! $required
+      header         TYPE zif_aff_types_v1=>ty_header_60,
+      "! <p class="shorttext">Text Index</p>
+      "! Standard- or text-Index
+      "! $required
+      "! $default 'abap_false'
+      text_index     TYPE abap_bool,
```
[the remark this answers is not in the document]
jheinri: That is in principle a good idea. The question is how this will look afterwards on the GUI. So far we are much in favor of a checkbox which only must be selected if the developer actually wants a text index (should be the 1% case). I guess if the change this to an enumeration we cannot have a checkbox, right?  
The question is also related to the support of old-style fulltext indexes. That would require the ability to enable and disable certain fields based on the index type. I think so far we said that this is not possible.

file-formats/tabl/type/zif_aff_indx_v1.intf.abap
```diff
@@ -0,0 +1,42 @@
...
+    "! Index fields
+    BEGIN OF ty_indexfield,
+      "! <p class="shorttext">Field Name</p>
+      "! Field name
+      fieldname TYPE c LENGTH 30,
+      "! <p class="shorttext">Field Position</p>
+      "! Position
+      position  TYPE n LENGTH 4,
+    END OF ty_indexfield,
+    tt_index_fields TYPE SORTED TABLE OF ty_indexfield WITH UNIQUE KEY fieldname.
```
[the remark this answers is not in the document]
jheinri: That would be at least against our DDIC coding conventions. We start simple type definitions with ty_, table types with tt_.  If you have other guidelines and it is important for you I have no problem if we change this, but from our perspective this looks ok as it is.

file-formats/tabl/type/zif_aff_indx_v1.intf.abap
```diff
@@ -0,0 +1,42 @@
...
+      "! <p class="shorttext">Text Index</p>
+      "! Standard- or text-Index
+      "! $required
+      "! $default 'abap_false'
+      text_index     TYPE abap_bool,
+      "! <p class="shorttext">Unique</p>
+      "! Unique flag
+      "! $required
+      "! $default 'abap_false'
+      unique_index   TYPE abap_bool,
```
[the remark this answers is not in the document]
jheinri: "UNIQUE" is a keyword is several languages, including SQL. We try to never use keywords as identifiers for something. Therefore I would leave it as it is.

file-formats/tabl/type/zif_aff_indx_v1.intf.abap
```diff
@@ -0,0 +1,42 @@
...
+      "! <p class="shorttext">Fields</p>
+      "! Index fields
+      index_fields   TYPE tt_index_fields,
```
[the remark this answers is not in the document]
jheinri: No, that is not possible. I will add the "$required" here.

file-formats/tabl/type/zif_aff_indx_v1.intf.json
```diff
@@ -0,0 +1,7 @@
+{
+  "formatVersion": "1",
+  "header": {
+    "description": "DDIC: form based editor for secondary indexes",
```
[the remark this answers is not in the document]
jheinri: Fine

file-formats/tabl/type/zif_aff_indx_v1.intf.abap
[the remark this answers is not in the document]
schneidermic0: Fine for me.

file-formats/tabl/type/zif_aff_indx_v1.intf.abap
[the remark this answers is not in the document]
schneidermic0: It would become a drop-down with both options. The "old-style fulltext indexes" would be a third option, wouldn't it?

For the ABAP file formats it is most probably helpful to specify the full object type even if we use/support just a subset in ADT in a first version.

file-formats/tabl/type/zif_aff_indx_v1.intf.abap
```diff
@@ -0,0 +1,42 @@
...
+  TYPES:
+    "! <p class="shorttext">ABAP DDIC Secondary Index Properties</p>
+    "! Object type INDX
+    BEGIN OF ty_main,
+      "! $required
+      format_version TYPE zif_aff_types_v1=>ty_format_version,
+      "! <p class="shorttext">Header</p>
+      "! Header
+      "! $required
+      header         TYPE zif_aff_types_v1=>ty_header_60,
```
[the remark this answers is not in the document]
wurzka: For REPS & FUNC, we use the header `zif_aff_types_v1=>ty_header_only_description`, a header for sub objects which consists only of the field `description`. Should this be kept the same?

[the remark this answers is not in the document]
jheinri: The original language can be derived from the table. That is not necessary true for the abap_language_version. THere are still discussion going on, but currently we must assume that extension indexes (R3TR XINX) can have an abap_labguage_version independent from the table. 
So far we assume that the file format for LIMU INDX and R3TR XINX is the same. If this is still true this would mean that we leave the header field as it is, correct?

larshp:
I think "b)" has more sub-options, as the source code for table definitions, I dont think it covers POOL/CLUSTER tables?

file-formats/tabl/type/zif_aff_indx_v1.intf.abap
jheinri: Very good question. The idea with the "..sorted table...with unique key field_name" was to ensure the uniqueness of each field name. But the index field order of course does not depend on the index field names. So the table cannot be ordered by field_name. 
The order of the index fields is very important. It is implicitly derived from the position of the fields in the index_fields table. 
-> I think we can get rid of the "position" field here as well. At least we do not like to see the "position" in a GUI editor.
-> Any ideas how we can ensure uniqueness of the index field names?

schneidermic0: > -> Any ideas how we can ensure uniqueness of the index field names?

What do you think about adding a secondary key to the internal table with a unique key field `field_name`
```
  TYPES
    tt_index_fields TYPE standard tABLE OF ty_indexfield WITH DEFAULT KEY
        WITH UNIQUE SORTED KEY name_key COMPONENTS field_name.
```

Adding a unique key to an internal table would mean that you cannot transform a JSON that has duplicate keys. The transformation will raise an exception

file-formats/tabl/type/zif_aff_indx_v1.intf.abap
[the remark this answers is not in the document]
schneidermic0: I discussed this with the team. We would prefer to stick only to one option within this repository. I.e., types shall start with `ty_`. 

See also #357

file-formats/tabl/type/zif_aff_indx_v1.intf.abap~
```diff
@@ -0,0 +1,42 @@
+INTERFACE zif_aff_indx_v1
+  PUBLIC.
```
larshp, line 2: bad filename, should end with `.abap` instead of `.abap~`

file-formats/tabl/type/zif_aff_indx_v1.intf.abap
```diff
@@ -0,0 +1,71 @@
...
+    END OF ty_indexfield,
+    tt_index_fields TYPE TABLE OF ty_indexfield WITH DEFAULT KEY
+      WITH UNIQUE SORTED KEY name_unique COMPONENTS field_name.
+
+  TYPES:
+    "! <p class="shorttext">ABAP DDIC Secondary Index Properties</p>
+    "! Object type INDX
+    BEGIN OF ty_main,
+      "! $required
+      format_version        TYPE if_aff_types_v1=>ty_format_version,
```
schneidermic0, line 21:
```suggestion
      format_version        TYPE zif_aff_types_v1=>ty_format_version,
```

file-formats/tabl/type/zif_aff_indx_v1.intf.abap
```diff
@@ -0,0 +1,71 @@
...
+      "! <p class="shorttext">Header</p>
+      "! Header
+      "! $required
+      header                TYPE if_aff_types_v1=>ty_header_60,
```
schneidermic0, line 25: Do we need a full-blown `header`? I guess the object needs only a description. Therefore, you might want to change the type to `zif_aff_types_v1=>ty_header_only_description`

```suggestion
      header                TYPE zif_aff_types_v1=>ty_header_only_description,
```

file-formats/tabl/type/zif_aff_indx_v1.intf.abap~
schneidermic0: I agree. I guess this file should be deleted, isn't it? There is already a file `zif_aff_indx_v1.intf.abap.

Also file `.zif_aff_indx_v1.intf.abap.un~` should be removed

file-formats/tabl/indx-v1.json
```diff
@@ -0,0 +1,109 @@
+{
```
schneidermic0, line 1: The json-schema file seems not to be updated, to the new interface version, yet.

schneidermic0 (changes requested):
Thanks for your update, @jheinri. 

Please see my and @larshp's comments and questions.

file-formats/tabl/type/zif_aff_indx_v1.intf.abap
```diff
@@ -0,0 +1,71 @@
...
+      "! Index fields
+      index_fields          TYPE tt_index_fields,
+      "
+      " the following are attributes for the fulltext index,
+      " available in HANA onPrem only
+      "
+      "! <p class="shorttext">Language Column for Full-Text Index</p>
+      fti_langu_column      TYPE c LENGTH 30,
+      "! <p class="shorttext">Column for Mime Type</p>
+      fti_mime_type_col     TYPE c LENGTH 30,
```
schneidermic0, line 47: Wouldn't it make sense to avoid abbreviations like `col`, `langu` or `indx` later and replace them by `column`, `language` or `index`?

schneidermic0 (changes requested):
Two additional comments related to full-text indexes. I think I missed to add them in my previous review.

file-formats/tabl/type/zif_aff_indx_v1.intf.abap
jheinri: As discussed, we need a full-blown header with abap_language_version. No change here.

file-formats/tabl/type/zif_aff_indx_v1.intf.abap
```diff
@@ -0,0 +1,71 @@
...
+      no_db_index           TYPE abap_bool,
+      "! <p class="shorttext">Fields</p>
+      "! Index fields
+      index_fields          TYPE tt_index_fields,
+      "
+      " the following are attributes for the fulltext index,
+      " available in HANA onPrem only
+      "
+      "! <p class="shorttext">Language Column for Full-Text Index</p>
+      fti_langu_column      TYPE c LENGTH 30,
```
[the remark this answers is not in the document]
jheinri: That's a good idea! I assume that even if the fields are in a sub-structure I can hide them in the editor...

file-formats/tabl/type/zif_aff_indx_v1.intf.abap
jheinri: In general probably yes, here in this context no. These are the column names that are used to store these FTI properties in a DDIC metadata table. As these names are already complicated, I would prefer to have a 1:1 mapping to the existing column names here and not change them again.

schneidermic0: Mmmh, but this is actually the intension of the ABAP file formats: Remove the complexity for developers (end users) working with the object type. They shouldn't know the field names on the DB table.

file-formats/tabl/type/zif_aff_indx_v1.intf.abap
schneidermic0: I agree, for R3TR XINX, you will need it. 
But for LIMU objects, I expect the ABAP language version and the original language are specified for the table.

I would create different headers for R3TR XINX and LIMU XINX

file-formats/tabl/type/zif_aff_indx_v1.intf.abap
jheinri: Still disagree. Someone needs to map these field names to the appropriate values in the ABAP DDIC. If this is no longer a one-to-one mapping, you add complexity instead of reducing it. I admit that this is a little bit "seen from my point of view"..

schneidermic0: Maybe, I get you wrong. Wouldn't there be still a 1:1 relation ship if we just rename an field in the file format? E.g., rename field `fti_langu_column` to something like `language_column`.

schneidermic0:
I close this PR. No updates for more than one year.

PR #361, opened 2022-05-05
[CHDO] Renaming an attribute

albertmink:
We decided to rename the attribute. Changes are already applied in dev system.

### DIFF
--- a/file-formats/chdo/examples/z_aff_example_chdo.chdo.json
+++ b/file-formats/chdo/examples/z_aff_example_chdo.chdo.json
@@ -38,6 +38,6 @@
     "number": "600"
   },
   "generatedObject": {
-    "classOrFunction": ""
+    "name": ""
   }
 }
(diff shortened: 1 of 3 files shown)


file-formats/chdo/examples/z_aff_example_chdo.chdo.json
```diff
@@ -38,6 +38,6 @@
     "number": "600"
   },
   "generatedObject": {
-    "classOrFunction": ""
+    "name": ""
```
albertmink, line 41: @wurzka is it allowed to have an empty entry? I mean from an ABAP perspective. (JSON Schema wise, it is fine!)
The deserialized object will have errors, right? The mandatory field is initial, so this a syntax error I guess.

wurzka: Mh, difficult to say. This entry is read-only and gets created by activating. So I'm not sure if it would result in a syntax error.
Maybe we should fill it with something like 
```suggestion
    "name": "ZCL_AFF_EXAMPLE_CHDO"
```

albertmink: I see, the activation process will hopefully provide the information.

schneidermic0 (approved):
LGTM

PR #376, opened 2022-07-11
Consider default values in ABAP file formant implementations

schneidermic0:
We just had a discussion that it is crucial to consider the default handling during serialization and deserialization. Otherwise, there might be a mix-up with initial values.

Let's update the documentation with this information.


wurzka (approved):
LGTM :)

albertmink (approved):
Thanks.

PR #390, opened 2022-08-31
Add extensions to GitHub Codespaces

albertmink:
Maybe we can simplify the contributions that reaches us via the GitHub Codespaces, as with some preinstalled extensions we could lower the entry hudle 🤔
➡️  no trailing whitespaces and EOF issues would be a good starting point


albertmink:
Requirement for extension to run in Codespaces [here ](https://code.visualstudio.com/api/extension-guides/web-extensions)
EditorConfig doesn't plan to provide this requirement, see https://github.com/editorconfig/editorconfig-vscode/issues/327

Maybe there are other extension to ensure EOL, EOF, indentation and trailing whitespaces.

albertmink:
After asking the org admin, we got the answer _... codespaces is not covered by our GitHub contract ..._

No Codespaces for now, maybe someday
image

PR #393, opened 2022-08-31
[WB2J] Add new object type 

wurzka:
Add files for demo object type wb2j


wurzka:
Pull request is closed since it was only used for demo purpose.

PR #400, opened 2022-09-28
[CHDO] add chdo files

TomK01:


schneidermic0 (changes requested):
Thanks @TomK01 for your contrubution. 👍 

Your changes look good to me. I think there is just one issue in the README. I have made a proposal which also should fix the markdown link error

schneidermic0 (approved):
Looks good to me.

Welcome as contributor. :)

PR #407, opened 2022-10-26
RONT branch

Ehret-Thomas:
Reworked ABAP File Format for object type RONT (SAP Object Type)


schneidermic0 (commented):
See my comment

schneidermic0 (approved):
Thanks for clarification.

Looks good to me. Thanks, @Ehret-Thomas.

PR #417, opened 2022-11-02
Update chko example

huber-nicolas:


schneidermic0 (approved):
LGTM

albertmink:
@huber-nicolas in #426 the markdown checker does not check anymore for the `API.REUSE....`. This could be a workaround to prevent this `false positives`

PR #422, opened 2022-11-02
update scp1

huber-nicolas:


wurzka (approved):
LGTM :)

huber-nicolas:
@schneidermic0 I just realised that this example is very big. It has 166 thousand rows. Do we need such a big example?

schneidermic0:
> @schneidermic0 I just realised that this example is very big. It has 166 thousand rows. Do we need such a big example?

I guess this we have to discuss with the object type owners. I guess we can reduce it. ;)

However, this should not hinder you merge this PR. This can be also updated in a separate PR.

schneidermic0 (approved):
LGTM

PR #439, opened 2022-11-23
DDLS: Use Official Terms for "sourceType" Attribute

BeckerWdf:
The official term for source type "v" is "DDIC-Based Views". So let's also use it here.

### DIFF
--- a/file-formats/ddls/ddls-v1.json
+++ b/file-formats/ddls/ddls-v1.json
@@ -105,7 +105,7 @@
       "description": "Source Type",
       "type": "string",
       "enum": [
-        "view",
+        "ddicBasedView",
         "viewEntity",
         "viewExtend",
         "viewEntityExtend",
@@ -117,7 +117,7 @@
         "projectionView"
       ],
       "enumTitles": [
-        "View",
+        "DDIC-Based Views",
         "View Entity",
         "View Extend",
         "View Entity Extend",
@@ -129,7 +129,7 @@
         "Projection View"
       ],
       "enumDescriptions": [
-        "View",
+        "DDIC-Based Views",
         "View Entity",
         "View Extend",
         "View Entity Extend",
--- a/file-formats/ddls/type/zif_aff_ddls_v1.intf.abap
+++ b/file-formats/ddls/type/zif_aff_ddls_v1.intf.abap
@@ -15,9 +15,9 @@ INTERFACE zif_aff_ddls_v1
     "! <p class="shorttext">Source Type</p>
     "! Source type
     BEGIN OF co_source_type,
-      "! <p class="shorttext">View</p>
-      "! View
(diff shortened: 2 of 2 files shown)


schneidermic0:
@BeckerWdf I understood you also wanted to rename the actual enum value "view", didn't you?

This would be an incompatible change, however since DDLS is not yet used, I think we can still change it.

file-formats/ddls/ddls-v1.json
```diff
@@ -129,7 +129,7 @@
         "Projection View"
       ],
       "enumDescriptions": [
-        "View",
+        "DDIC-Based Views",
         "View Entity",
         "View Extend",
         "View Entity Extend",
```
schneidermic0, line 135: Not really  related to the actual PR. But I see descriptions are title style. They shall be sentence style. This is true for the whole DDLS file format.

```suggestion
        "DDIC-based views",
        "View entity",
        "View extend",
        "View entity extend",
```

@BeckerWdf Do you want to update all descriptions in this PR or shall I create an separate issue?

file-formats/ddls/type/zif_aff_ddls_v1.intf.abap
```diff
@@ -15,9 +15,9 @@ INTERFACE zif_aff_ddls_v1
...
     "! Source type
     BEGIN OF co_source_type,
-      "! <p class="shorttext">View</p>
-      "! View
-      view               TYPE ty_source_type VALUE 'V',
+      "! <p class="shorttext">DDIC-Based Views</p>
+      "! DDIC-Based Views
+      ddic_based_view    TYPE ty_source_type VALUE 'V',
       "! <p class="shorttext">View Entity</p>
       "! View Entity
```
schneidermic0, line 22:
```suggestion
      "! DDIC-based Views
      ddic_based_view    TYPE ty_source_type VALUE 'V',
      "! <p class="shorttext">View Entity</p>
      "! View entity
```

schneidermic0 (changes requested):
Thank you,  @BeckerWdf, for updating the file format. 👍 

In general the changes are fine, because we can still change incompatible for DDLS.

I just found a minor issue related to the descriptions in general. See comment below

file-formats/ddls/ddls-v1.json
BeckerWdf: @schneidermic0: Let's handle this in a separate issue.

schneidermic0: I have just created a new issue: #441

file-formats/ddls/type/zif_aff_ddls_v1.intf.abap
schneidermic0: See #441

schneidermic0 (approved):
We fix descriptions with #441. All other changes look fine to me.

PR #446, opened 2022-12-01
DDLS: Service Consumption Model is a name

### LINKED ISSUE #443: DDLS: Service Consumption Model is a name
In the cause of #441 we missed that "Service Consumption Model" is a name and therefore always to be capitalized (just like "ABAP Development Tools")

BeckerWdf:
As it's a name it's also capitalized even in descriptions / sentences.

Fixes #443


schneidermic0 (approved):
Thanks, @BeckerWdf.

Looks good to me.

PR #447, opened 2022-12-01
Update zif_aff_evtb_v1.intf.abap

apanys:
Hi colleagues,
as internally discussed, we change the type in a compatible way from the AFF type to plain C 30.
Kind regards
André


schneidermic0 (approved):
Thanks, @apanys. 

Looks good to me.

PR #451, opened 2022-12-07
Feature/gsmp

mseich:
Add Instance Creation Mode

### DIFF
--- a/file-formats/gsmp/type/zif_aff_gsmp_v1.intf.abap
+++ b/file-formats/gsmp/type/zif_aff_gsmp_v1.intf.abap
@@ -24,7 +24,7 @@ INTERFACE zif_aff_gsmp_v1
     "! <p class="shorttext">Instantiation</p>
     "! Instantiation
     "! $values {@link zif_aff_gsmp_v1.data:co_instantiation}
-    "! $default {@link zif_aff_gsmp_v1.data:co_instantiation.create}
+    "! $default {@link zif_aff_gsmp_v1.data:co_instantiation.create_new}
     ty_instantiation TYPE c LENGTH 1,
 
     "! <p class="shorttext">Implementation</p>
@@ -150,7 +150,7 @@ INTERFACE zif_aff_gsmp_v1
 
       "! <p class="shorttext">Creating New Instances</p>
       "! Creating new instances
-      create      TYPE ty_instantiation VALUE 'C',
+      create_new  TYPE ty_instantiation VALUE 'C',
 
     END OF co_instantiation.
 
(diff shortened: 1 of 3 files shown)


schneidermic0:
Just a heads-up. We have a similar property (like `instanceCreationMode`) for BAdI definitions:

https://github.com/SAP/abap-file-formats/blob/c83d2cfecc48ad77049986d52477179638dfe5ae/file-formats/enhs/enhs-v1.json#L119-L138

Would it make sense to use the same values for both?

wurzka (changes requested):
Looks good to me, but could you also please update the example for GSMP? It should be enough to add the line 
`"instanceCreationMode": "create",`
between those
https://github.com/SAP/abap-file-formats/blob/652d7830a4ccac0980c66d2a5b1f07884297f221/file-formats/gsmp/examples/z_aff_example_gsmp.gsmp.json#L13-L14

schneidermic0:
Addition to my comment above with a proposal to use the same terms in GSMP and ENHS.

We could use the field `instantiation` for GSMP and ENHS with the values `createNew` and `reuse` (for BAdIs we add `contextSpecific`). 

I have no clear opinion about the names. Other proposals are welcome.

@mseich @marcushoepfner What do you think?

mseich:
createNew and reuse would be OK for me. I'll discuss this with the team. Eventually we will add possible values to the instantiation field later like 'adaptive'. In that case ENHS and GSMP would drift apart anyway.

file-formats/gsmp/type/zif_aff_gsmp_v1.intf.abap
```diff
@@ -127,7 +138,21 @@ INTERFACE zif_aff_gsmp_v1
...
+    "! Instantiation
+    BEGIN OF co_instantiation,
+
+      "! <p class="shorttext">Reuse Instances</p>
+      "! Reuse instances
+      reuse       TYPE ty_instantiation VALUE 'R',
+
+      "! <p class="shorttext">Creating New Instances</p>
+      "! Creating new instances
+      create      TYPE ty_instantiation VALUE 'C',
```
wurzka, line 153: Since we decided on 'createNew', shouldn't it be 
```suggestion
      create_new      TYPE ty_instantiation VALUE 'C',
```
?

schneidermic0:
> @marcushoepfner What do you think?

I have talked to @marcushoepfner earlier today. He agrees to adapt the enum values for ENHS

schneidermic0 (approved):
Looks good to me.

Thank you @mseich for keeping the file format up-to-date. 👍

wurzka (approved):
Thanks, looks good to me!

PR #452, opened 2022-12-07
Update GitPython dependency to latest version

schneidermic0:
Can we update the dependency to GitPython the latest version? 

Fixes https://github.com/SAP/abap-file-formats/security/dependabot/1

We would need to update to at least version 3.1.20


albertmink (approved):
LGTM

PR #469, opened 2023-01-13
[NTTY] Add new object type NTTY

thomas-kaltenbach:


file-formats/ntty/type/zif_aff_ntty_v1.intf.abap
```diff
@@ -0,0 +1,70 @@
...
+    "! General information
+    BEGIN OF ty_general_information,
+      "! <p class="shorttext">Language Dependent</p>
+      "! Language dependent
+      "! $showAlways
+      language_dependency TYPE abap_bool,
+      "! <p class="shorttext">Multiple Notes</p>
+      "! Multiple notes possible
+      "! $showAlways
+      multiple            TYPE abap_bool,
```
wurzka, line 24: What do you think about
```suggestion
      multiple_notes            TYPE abap_bool,
```
?

file-formats/ntty/type/zif_aff_ntty_v1.intf.abap
```diff
@@ -0,0 +1,70 @@
...
+      "! <p class="shorttext">Hide Title</p>
+      "! Hides the title
+      no_subject          TYPE abap_bool,
```
wurzka, line 27: Here, title/description are very different from the field name. What about 
```suggestion
      hide_title          TYPE abap_bool,
```
?

file-formats/ntty/type/zif_aff_ntty_v1.intf.abap
```diff
@@ -0,0 +1,70 @@
...
+  PUBLIC.
+
+
+  TYPES:
+    "! $values { @link zif_aff_ntty_v1.data:co_profile }
+    "! $default { @link zif_aff_ntty_v1.data:co_profile.text }
+    ty_profile TYPE c LENGTH 20,
+
+    "! $minimum 0
+    "! $maximum 2147483647
```
wurzka, line 11: As type is of type I, you don't need this annotation for maximum. Our schema generator adds the maximum automatically to the JSON Schema. It is only needed if extrema differ from the value range of the built in ABAP type
```suggestion
```

file-formats/ntty/type/zif_aff_ntty_v1.intf.abap
```diff
@@ -0,0 +1,70 @@
...
+
+    "! <p class="shorttext">Content Settings</p>
+    "! Content settings
+    BEGIN OF ty_content_settings,
+      "! <p class="shorttext">Maximum Length</p>
+      "! Maximum length
+      max_length TYPE ty_max_length,
+      "! <p class="shorttext">Formatting Profile</p>
+      "! Formatting profile
+      profile    TYPE ty_profile,
```
wurzka, line 38: What about
```suggestion
      formatting_profile    TYPE ty_profile,
```
?

file-formats/ntty/type/zif_aff_ntty_v1.intf.abap
```diff
@@ -0,0 +1,70 @@
...
+      "! Hides the title
+      no_subject          TYPE abap_bool,
+    END OF ty_general_information,
+
+    "! <p class="shorttext">Content Settings</p>
+    "! Content settings
+    BEGIN OF ty_content_settings,
+      "! <p class="shorttext">Maximum Length</p>
+      "! Maximum length
+      max_length TYPE ty_max_length,
```
wurzka, line 35: Maybe also
```suggestion
      maximum_length TYPE ty_max_length,
```
?

file-formats/ntty/type/zif_aff_ntty_v1.intf.abap
```diff
@@ -0,0 +1,70 @@
...
+      content_settings    TYPE ty_content_settings,
+    END OF ty_main.
+
+  CONSTANTS:
+    BEGIN OF co_profile,
+      "! <p class="shorttext">Text Only</p>
+      "! Plain text content (text/plain)
+      text      TYPE ty_profile VALUE 'TEXT',
+      "! <p class="shorttext">Simple Formatting</p>
+      "! simple formatted content (text/html)
```
wurzka, line 66:
```suggestion
      "! Simple formatted content (text/html)
```

file-formats/ntty/type/zif_aff_ntty_v1.intf.abap
```diff
@@ -0,0 +1,70 @@
...
+      rte_basic TYPE ty_profile VALUE 'RTE_BASIC',
```
wurzka, line 67: Maybe something like
```suggestion
      simple_formatting TYPE ty_profile VALUE 'RTE_BASIC',
```
? It is more human readable.

wurzka (changes requested):
Thanks for your contribution @thomas-kaltenbach.
Please wait with adapting my suggestions until we get one more review. Maybe I will be outvoted.

file-formats/ntty/type/zif_aff_ntty_v1.intf.abap
```diff
@@ -0,0 +1,70 @@
...
+    "! Notetype
+    BEGIN OF ty_main,
+      "! <p class="shorttext">Format Version</p>
+      "! Format version
+      "! $required
+      format_version      TYPE zif_aff_types_v1=>ty_format_version,
+      "! <p class="shorttext">Header</p>
+      "! Header
+      "! $required
+      header              TYPE zif_aff_types_v1=>ty_header_60,
```
wurzka, line 51: Does your object type support also `"keyUser"` as abap language version? If yes, this header is fine. Otherwise, it would be `ty_header_60_cloud`

file-formats/ntty/type/zif_aff_ntty_v1.intf.abap
schneidermic0: "html" might be also an option if this flag indicates that the text is content type (text/html)

file-formats/ntty/type/zif_aff_ntty_v1.intf.abap
```diff
@@ -0,0 +1,70 @@
...
+      "! <p class="shorttext">General Information</p>
+      "! General information
+      general_information TYPE ty_general_information,
```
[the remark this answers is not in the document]
schneidermic0: > Would it be better to have the three bools on this level without an extra structure?

I think we came to the conclusion that it is easier to handle if there is a dedicated structure. If we have a specific/meaningful name for the section, it would be the best. For others, I think `general_information` is fine.

> Maybe, @schneidermic0, we should think about a general suggestion.

I agree

file-formats/ntty/type/zif_aff_ntty_v1.intf.abap
```diff
@@ -0,0 +1,70 @@
...
+    "! $minimum 0
+    "! $maximum 2147483647
+    ty_max_length TYPE i,
+
+    "! <p class="shorttext">General Information</p>
+    "! General information
+    BEGIN OF ty_general_information,
+      "! <p class="shorttext">Language Dependent</p>
+      "! Language dependent
+      "! $showAlways
```
schneidermic0, line 19: Just a remark: I would try to avoid `$showAlways` in booleans if `false` is the default. This would shrink the JSON file without loosing information.

schneidermic0 (commented):
See my comments

file-formats/ntty/type/zif_aff_ntty_v1.intf.abap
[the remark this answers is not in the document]
thomas-kaltenbach: We've already looked for a good name but couldn't find one, so we checked other objects and found the term "general_information". 
It sounds like that's ok, so I'd leave it as it is.

file-formats/ntty/type/zif_aff_ntty_v1.intf.abap
thomas-kaltenbach: I removed the `$showAllways`. Thank you for the hint.

file-formats/ntty/type/zif_aff_ntty_v1.intf.abap
thomas-kaltenbach: I changed it to `ty_header_60_cloud`. Thank you for the hint.

thomas-kaltenbach:
Hi @wurzka,
thank you for the review! 
I applied all the suggestions, it makes sense for me.

file-formats/ntty/type/zif_aff_ntty_v1.intf.abap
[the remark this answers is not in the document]
wurzka: We talked in our team about this issue. We recommend using `general_information` as you did. So this is fine for us.

wurzka (approved):
Thanks for your changes. Now, it looks good to me.

schneidermic0 (approved):
@thomas-kaltenbach Welcome as contributor! 👍

PR #478, opened 2023-02-03
[RONT] Add default annotation for field type category.

Ehret-Thomas:
Specify "Business Object" as the default type category for object type RONT (SAP Object Type).


schneidermic0 (approved):
Compatible change, since the field was an enum and (even) required.

Thanks @Ehret-Thomas for the update.

huber-nicolas (approved):

Issue #479, opened 2023-02-06
GH Action: JSON data validation is not working properly


albertmink:
The JSON data is obviously not valid for the given JSON Schema, see 'formatVersion' that is required but not provided
https://github.com/SAP/abap-file-formats/blob/63a267d6e681ae452220c4cf7c1625ae36528ab7/file-formats/ntty/examples/z_aff_example_ntty.ntty.json#L1

albertmink:
Object types containing other characters than [a-z] are not covered, e.g.
https://github.com/SAP/abap-file-formats/tree/main/file-formats/scp1

albertmink:
Todo:
Algorithm might loop over JSON Schema and then pick the JSON data. As it is more likely that the JSON Schema is correct, than the JSON data. As of now, if the JSON data is corrupt, e.g. missing formatVersion, then the corresponding JSON Schema cannot be calculated. Hence, there will be no error reported.

albertmink:
Reopened because the `todo` above.

schneidermic0:
The check due to #480 seems to work. However, the error message is hard to understand. 

I suggest to change the error message to "JSON Example does not specify the field `formatVersion`".

Additionally, I would rename the check from "Validate JSON data" to "Validate JSON examples.

I'll create a PR

PR #480, opened 2023-02-06
Fix the JSON data validation

albertmink:
workaround for #479 that improves the validation and fixes the JSON data by educated guess.


schneidermic0 (approved):
Thanks, @albertmink!

LGTM

Issue #370, opened 2022-06-28
Upload example object for SPRV


huber-nicolas:
We will merge the pull request #369 without the example object. We have to upload the example object later (in a new pull request).

PR #482, opened 2023-02-07
Make validation of JSON data more robust

### LINKED ISSUE #479: GH Action: JSON data validation is not working properly

albertmink:
closed #479 

Logic: For every JSON Schema we find the JSON data/example and create a match. All matches then, are validated.
Needs some rework, once there will be other formatVersions than `1` :)

### DIFF
--- a/json-validator.py
+++ b/json-validator.py
@@ -8,41 +8,8 @@ import sys
 import re
 
 msg_errors = list()
-schemas = glob.glob('./file-formats/*/*.json')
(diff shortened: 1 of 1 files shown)


schneidermic0 (approved):
LGTM

Now, the validation reports a message for missing example file which is also reported in #370.

PR #483, opened 2023-02-13
Define Structuring of ty_main

### LINKED ISSUE #472: Structuring of ty_main - General Information
Each object type in this repository has a type definition, i.e., an ABAP type that contains all the information needed for the ABAP File Format. For each object type, this ABAP type is a structure, called `ty_main`. 

This structure can have several components which can be structures, too, but also tables, strings, numbers, booleans,....
In this issue, I want to focus on the components of `ty_main` that are neither structures nor tables. I will call them "simple components" in the following.  

For object type INTF, for example, these simple components are `category` and `proxy`.
https://github.com/SAP/abap-file-formats/blob/e7880e6c87637d50f6d750b6fedbf025f6b58c25/file-formats/intf/type/zif_aff_intf_v1.intf.abap#L45-L50

For CHKO, we have `category`, implementing_class` and `check_type`.
https://github.com/SAP/abap-file-formats/blob/e7880e6c87637d50f6d750b6fedbf025f6b58c25/file-formats/chko/type/zif_aff_chko_v1.intf.abap#L40-L50

In internal discussions, we agreed that we do not want to have such simple components on top level in `ty_main`. We agreed that all components of `ty_main` shall be structures or tables.

So the next question is how to name the structure where the former simple components shall be put into.
In case these simple components semantically belong together, it would be best to name the new component/structure accordingly. 

If not, we would suggest to call this component `general_information` like it is already done for CHDO.
https://github.com/SAP/abap-file-formats/blob/284683bd1417871c821500d62682847b7488adae/file-formats/chdo/type/zif_aff_chdo_v1.intf.abap#L122-L124

wurzka:
Closes #472

### DIFF
--- a/docs/json.md
+++ b/docs/json.md
@@ -73,6 +73,9 @@ The JSON Schema is generated based on the fields and their ABAP type specificati
 Fields `format_version` and `header` are mandatory and translate to `formatVersion` and `header` in the JSON Schema.
 The interface [`zif_aff_types_v1`](../file-formats/zif_aff_types_v1.intf.abap) offers a type for the format version and different headers for reuse, but also other often repeated types.
 
+Apart from the mandatory fields mentioned above, it is recommended to add only components to the structure `ty_main` that are either structures as well or tables. 
+A subcomponent consisting of general information about the object should be named `general_information` as far as no better name fits.
+
 To add more information to the JSON Schema than that provided by the ABAP type, ABAP Doc can be used.
 The comments are placed directly above the components of the type `ty_main`, but they are also read over several levels, e.g., in the case of nested structures.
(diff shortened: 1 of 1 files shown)


schneidermic0 (commented):
Looks good to me. Thanks :)

albertmink:
Are there checks/warnings panned in the report https://github.com/SAP/abap-file-formats-tools to notify any violation? Better have some automatism in place to ensure quality, than relying on the patient and wise reviewer.

wurzka:
There are no checks, yet, but I agree with you, @albertmink. I already thought about adding some warning or info messages to the generator.

PR #485, opened 2023-02-16
NTTY: implemented UX review hints

thomas-kaltenbach:


file-formats/ntty/type/zif_aff_ntty_v1.intf.abap
```diff
@@ -35,8 +35,8 @@ INTERFACE zif_aff_ntty_v1
       formatting_profile    TYPE ty_profile,
     END OF ty_content_settings,
 
-    "! <p class="shorttext">Notetype</p>
-    "! Notetype
+    "! <p class="shorttext">Note Type</p>
```
albertmink, line 38: Afaik the title here does not relate to any UI text, so you might put smth like `ABAP file format for Note Type objects` that acts as the title of the JSON Schema itself.

albertmink (changes requested):
Thank you for the adoptions, see comments.

file-formats/ntty/type/zif_aff_ntty_v1.intf.abap
thomas-kaltenbach: doen ;)

file-formats/ntty/type/zif_aff_ntty_v1.intf.abap
```diff
@@ -19,8 +19,8 @@ INTERFACE zif_aff_ntty_v1
       "! <p class="shorttext">Multiple Notes</p>
       "! Multiple notes possible
       multiple_notes            TYPE abap_bool,
-      "! <p class="shorttext">Hide Title</p>
-      "! Hides the title
+      "! <p class="shorttext">Title hidden</p>
```
albertmink, line 22: Another Title case, right @thomas-kaltenbach ?
```suggestion
      "! <p class="shorttext">Title Hidden</p>
```

file-formats/ntty/type/zif_aff_ntty_v1.intf.abap
```diff
@@ -19,8 +19,8 @@ INTERFACE zif_aff_ntty_v1
...
+      "! Title hidden
       hide_title          TYPE abap_bool,
```
albertmink, line 24: In order to align the ABAP variable and JSON field name to the UI text in the editor, you might consider to rename here accordingly. A simple rename in ADT should do the work for you ;)

```suggestion
      title_hidden          TYPE abap_bool,
```

See, the checkboxes in the UI 'Language Dependent' and 'Multiple Notes' correspond to 'language_dependent' (abap) or 'languageDependent' (json example). Same for 'Multiple Notes' but not for 'Title Hidden'.

To be clear, this is not a blocking issue im my opinion, but definitively a worthy change now.

file-formats/ntty/type/zif_aff_ntty_v1.intf.abap
thomas-kaltenbach: done

file-formats/ntty/type/zif_aff_ntty_v1.intf.abap
thomas-kaltenbach: Thank you for the hint. I also noticed it and if I change it I would also need to change it in the database table. Otherwise I have to map it everytime. This is unfortunately not only a rename with ADT :(

schneidermic0 (commented):
Same minor question as in the other PR. Is "note type" written in lower or upper case. In SAPterm I see (so far) only lower case variants.

thomas-kaltenbach:
I changed it to lower case. We have no plans to add it to SAPterm.

schneidermic0 (approved):
Thanks for the update 👍

PR #492, opened 2023-02-20
SAP BTP 2302 updates for SMBC file format

sepp4me:
Note: the commit text is incorrect, it should be 2302


wurzka (approved):
Thanks for your update! Looks good to me.

PR #495, opened 2023-02-22
[ENHS] Set enum defaults for compatibility

wurzka:


wurzka:
@schneidermic0, I went through your proposals with @marcushoepfner and added his suggestions. Could you double check it, please?

schneidermic0 (approved):
LGTM

PR #497, opened 2023-02-27
[UIST] Add new object type UIST

ViktoriaFreidel:

PR #507, opened 2023-03-10
Add header for cloud with description 100

wurzka:
Added new header that allows descriptions of length 100 and has ABAP Language Version "standard" and "cloudDevelopment"


schneidermic0:
In general, it looks good to me. 

Shall we add the additional type in #497. Then we have directly a usage of the type and don't need to change the configuration of abaplint for a short period of time.

wurzka:
Yes, I agree. I will close this PR without merging and add the header there

PR #510, opened 2023-03-17
WIP: Idea for handling of translations: Using same JSON structure for translation files

schneidermic0:
See #106. 

This PR shall show one possible approach how translations could be stored in separate JSON files (one file per language and JSON file). 

The filename pattern would be `<object name>.<object type>.<language>.json`.

The content of the JSON file shows following fields of the original ABAP file format (JSON representation):

- `formatVersion`
- all language-dependent fields
- all key fields to identify an item in an array

### DIFF
--- a/file-formats/chko/examples/z_aff_example_chko.chko.de.json
+++ b/file-formats/chko/examples/z_aff_example_chko.chko.de.json
@@ -1,0 +1,12 @@
+{
+  "formatVersion": "1",
+  "header": {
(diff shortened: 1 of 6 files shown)


schneidermic0:
Keys for arrays could be determined from `... TABLE OF ... WITH ... KEY ...` definitions of AFF ABAP type.

We need to mark fields which are language-dependent. Open question: Is this information also part of JSON-schema

schneidermic0:
We want to take the approach of .properties files. Therefore, I close this draft

PR #516, opened 2023-03-22
[EVTB] Refine ABAP Doc comments

Ehret-Thomas:
Integrating feedback from various teams the ABAP Doc comments in the AFF Interface for EVTB were refined. 
This has the effect that field labels in the native ADT editor for EVTB (Event Binding) are updated as well.

The following updates were made:

- The word "Event" is now omitted in the version naming, e.g. "Event Patch Version" --> "Patch Version", thus making the version naming more similar to the version naming of service bindings (SRVB).
- "Namespace" was refined to "Type Namespace" making it easier to see the connection between the namespace and the calculated type shown in the ADT UI.
- "Business Object" --> SAP Object to reflect that in almost all cases an SAP Object Type will be referenced here
- "Business Operation" was renamed to the more general term "Operation"
- Finally, moving from "Events" to "Event Versions" in the header for the event versions table is just a bit more precise.


schneidermic0:
@Ehret-Thomas Thanks for the update. 

Do you think it make sense to even update the field names? E.g., field `boName` is called "SAP Object Type", now. However, this would mean we have to increase the format version, because it is an incompatible change.

Ehret-Thomas:
@schneidermic0 thanks for the review and feedback! We are still discussing the two options in our team and will provide an update here as soon as we have reached a decision.

Ehret-Thomas:
After discussing the matter in our team, we decided to first update only the titles and labels with this pull request. As they are more precise now, we think these changes are a small improvement.
We plan to update the field names as well, but that update would come in a second version of the ABAP File Format for Event Binding.

schneidermic0 (approved):

wurzka (approved):

PR #518, opened 2023-03-24
Add documentation for content media type and encoding

wurzka:
With https://github.com/SAP/abap-file-formats-tools/pull/231, we added annotations for content media type and content encoding. 
This pull request adds the corresponding documentation.


wurzka:
Will be documented in https://github.com/SAP/abap-file-formats-tools since content encoding an d content media type should not be part of the AFF

PR #525, opened 2023-04-14
WIP: Idea for handling of translations: Using PO files

schneidermic0:
See https://github.com/SAP/abap-file-formats/issues/106

This PR shall show one possible approach how translations could be stored in separate .po files (one file per language) as suggested by @sbcgua in https://github.com/abapGit/abapGit/issues/2539.

The filename pattern would be `<object name>.<object type>.<language>.po`.

The keys are the texts to be translated.

Example:

```
#: object.array[arrayKey].field
msgid "Text in original language"
msgstr "Translated text"
```


schneidermic0:
We want to take the approach of .properties files. Therefore, I close this draft

PR #529, opened 2023-05-03
[SAJT] Move general attributes to new structure 'General Information'

heichmann:


schneidermic0:
As far as I know, we can still change the file format of SAJT incompatible. There is no use case yet (e.g., abapGit) that stores the file format outside of an ABAP server

schneidermic0 (approved):
@heichmann Thank you for the update. Looks good to me

PR #531, opened 2023-05-09
[RONT] Delete expanded name

Ehret-Thomas:
The expanded name has become obsolete and is therefore removed from the AFF Interface of object type RONT.


albertmink (approved):
LGTM

schneidermic0 (approved):
Thanks, @Ehret-Thomas

huber-nicolas (approved):

huber-nicolas:
> I think we can still change this object type incompatible as it is not used in repositories (such as abapGit), yet.

I think so too. I will merge this pull request

PR #526, opened 2023-04-25
[SUSI] Add new object type SUSI

WDFYvonne:

PR #534, opened 2023-05-15
Catch exception if json schema not found

albertmink:
Will provide a message, instead of abortion, for the error in #526 (example name missing object type).


schneidermic0 (approved):
Thanks, @albertmink. LGTM

PR #535, opened 2023-05-25
Rename field root_node_flag and delete obsolete expanded_name

Ehret-Thomas:
* the boolean root_node_flag field is renamed to root_node in the AFF interface.
* the obsolelete expanded name is removed from the ABAP File Format for SAP Object Node Type (NONT)


wurzka (approved):
@Ehret-Thomas, thank you for updating the file format of NONT
Looks good to me!

PR #538, opened 2023-06-26
[SAJC]: Add Parameters, Groups, Sections

heichmann:


file-formats/sajc/type/zif_aff_sajc_v1.intf.abap
```diff
@@ -37,20 +50,105 @@ INTERFACE zif_aff_sajc_v1
...
+      indent               TYPE abap_bool,
+      "! <p class="shorttext">Mandatory</p>
+      "! Flag indicating whether the parameter is mandatory
+      mandatory            TYPE abap_bool,
+      "! <p class="shorttext">Hidden</p>
+      "! Flag indicating whether the parameter is hidden
+      hidden               TYPE abap_bool,
+      "! <p class="shorttext">Read Only</p>
+      "! Flag indicating whether the parameter is read only
+      read_only            TYPE abap_bool,
```
schneidermic0, line 93: Minor: Is the order of the fields important? If it is not important, I would consider to group all boolean values and put them to the end of the structure.

schneidermic0 (commented):
Looks quite good to me. I have just some minor comments so far

file-formats/sajc/type/zif_aff_sajc_v1.intf.abap
```diff
@@ -37,20 +50,105 @@ INTERFACE zif_aff_sajc_v1
...
       "! Name of the class which contains the notification exit
       notification TYPE zif_aff_types_v1=>ty_object_name_30,
+      "! <p class="shorttext">Delete</p>
+      "! Name of the class which contains the delete exit
+      delete       TYPE zif_aff_types_v1=>ty_object_name_30,
     END OF ty_exit_classes.
 
+  "! <p class="shorttext">Parameter Name</p>
+  "! Name of the parameter
+  TYPES ty_parameter_name TYPE c LENGTH 38.
```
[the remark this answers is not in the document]
schneidermic0: Do all values on DB have a length of 8 characters or less? If this was the case, I would set it to 8 chars. Otherwise I would keep it to the length stored on the database.

> The parameter name is set to 'read-only' in the configuration file, so they are only displayed, the customer cannot change them.

Keep in mind. You can avoid changing the length in a dedicated editor like in ADT. However, users can change it in the file (e.g., in a git repository when using abapGit).

file-formats/sajc/type/zif_aff_sajc_v1.intf.abap
schneidermic0: I agree, grouping would be nice in this case. I have to check this with the team.

file-formats/sajc/type/zif_aff_sajc_v1.intf.abap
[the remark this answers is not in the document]
heichmann: In a class-based catalog entry (these are the only ones which currently can be created and changed in the ADT) the length is restricted to 8 characters. But currently it is also possible to display catalog entry chains in the ADT (only display, no change). Here the length can be up to 38 characters. So, I think that I leave it to 38 characters.

It is no problem, if invalid entries are transferred via ABAPGit. The ADT only displays the parameters of the catalog entry which are defined in the ABAP class. If a parameter of the catalog entry is not defined in the ABAP class (e.g. because the class was changed meanwhile, or because some garbage is transferred via ABAPGit), the parameter is automatically removed from the parameter list (which is displayed in the ADT; to remove it from the database, the user must save and activate). An information message is displayed which informs the user about the removal and the reason (parameter does not exist in class).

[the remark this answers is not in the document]
schneidermic0: > In a class-based catalog entry (these are the only ones which currently can be created and changed in the ADT) the length is restricted to 8 characters. But currently it is also possible to display catalog entry chains in the ADT (only display, no change). Here the length can be up to 38 characters. So, I think that I leave it to 38 characters.

If there are values with 38 characters. We should stay with 38 characters.

file-formats/sajc/type/zif_aff_sajc_v1.intf.abap
```diff
@@ -37,20 +50,105 @@ INTERFACE zif_aff_sajc_v1
...
-      "! Name of the program which is executed within the job
-      program_name TYPE ty_program_name,
-    END OF ty_general_information.
+      name                 TYPE ty_parameter_name,
+      "! <p class="shorttext">Group</p>
+      "! Name of the parameter group
+      group                TYPE ty_group_name,
+      "! <p class="shorttext">Indent</p>
+      "! Flag indicating whether the parameter is indented on the selection screen
+      indent               TYPE abap_bool,
```
wurzka, line 84: We suggest using 
```suggestion
      indented              TYPE abap_bool,
```

file-formats/sajc/type/zif_aff_sajc_v1.intf.abap
```diff
@@ -69,5 +167,14 @@ INTERFACE zif_aff_sajc_v1
...
       exit_classes        TYPE ty_exit_classes,
+      "! <p class="shorttext">Parameters</p>
+      "! List of parameters
+      parameters          TYPE ty_parameter_table,
+      "! <p class="shorttext">Groups</p>
+      "! List of groups
+      groups              TYPE ty_group_table,
+      "! <p class="shorttext">Sections</p>
+      "! List of sections
+      sections            TYPE ty_section_table,
```
wurzka, line 178: We suggest changing the order:
1. Sections
2. Groups
3. Parameters

wurzka (changes requested):
UX review has been performed. See my comments

file-formats/sajc/type/zif_aff_sajc_v1.intf.abap
heichmann: I changed it according to the suggestion

file-formats/sajc/type/zif_aff_sajc_v1.intf.abap
heichmann: I changed it according to the suggestion

file-formats/sajc/type/zif_aff_sajc_v1.intf.abap
```diff
@@ -24,6 +24,19 @@ INTERFACE zif_aff_sajc_v1
   "! Name of the program
   TYPES ty_program_name TYPE c LENGTH 40.
 
+  TYPES:
+    "! <p class="shorttext">General Information</p>
+    "! General information
+    BEGIN OF ty_general_information,
+      "! <p class="shorttext">Class with Execute Method</p>
+      "! Name of the class which contains the execute-method to run within the job
+      "! $required
```
schneidermic0, line 33: Is`class_name` required? Isn't it sufficient to specify a `program_name`?

heichmann: Some words about the background: There are two different types of catalog entries: class based (which run an ABAP class in background) and report based (which run an ABAP report in background). Because in Steampunk only classes can be created, the ADT only allows to edit class based catalog entries.

Now the problem is that both types use the same transport object, so both types are visible on the left side of the ADT in the tree. This means that it is possible to select a report based catalog entry in the tree. I implemented this in a way that a report based catalog entry uses a different configuration file which sets everything to read-only.

Now there are two problems:
- In a report based catalog entry no class name with title 'Class name' should be displayed, but a report name with title 'Program name'.
- The check method should display an error message which explains that report based catalog entries are not supported by the ADT. So, the check method should recognize that currently a report based catalog entry is displayed.

To realize this I used the following way:
- In a class based catalog entry the field class_name is filled (which is required) and only this field is visible on the screen.
- In a report based catalog entry, the field program_name is filled (which is read-only like everything else) and only this field is visible on the screen. 
- The check method checks whether the field program_name is filled and outputs an error message in this case.

Meanwhile there would be a better solution:
- I think the title of the field can be overwritten in the configuration file, can't it?
- Meanwhile the check method reads the catalog entry from the database (to check whether the user changed the class name and needs to sync the parameter list). So, the check method may also read the information from the database whether the catalog entry is report based. It does not need the field program_name any more,

But I don't think that I can remove the field program_name because of ABAPGit and incompatible change. I don't want to increase the version number because of this.

schneidermic0: Having both fields is fine for me. I was wondering whether the field `class_name` should be a `$required` field. Because obviously, it is not required, if there is a program.

I would like to double-check with @wurzka.

schneidermic0: I have opened an issue. See #546.

Maybe, we recommend to increase the format version later to solve the issue. However, I would keep it as it is, for now.

schneidermic0 (approved):

wurzka (approved):
LGTM :)

PR #545, opened 2023-07-26
Update documentation

albertmink:
* State the as-of-now regarding translation: `.properties` is for text elements
* Clarify that the metadata file `.json` is always there, regardless of the object type.


schneidermic0 (approved):
Looks good to me.

PR #548, opened 2023-08-10
[UIST] Update: Replace Base Template Id with Merge Id

ViktoriaFreidel:
We replace a Field for Base Template ID wit a Merge ID. This is only one correction.


schneidermic0:
Can be still changed incompatible without updating the format version

schneidermic0 (approved):
Thanks for the update, @ViktoriaFreidel.

If "merge ID" is the right terminology, it looks good to me.

file-formats/uist/examples/z_aff_example_uist.uist.json
```diff
@@ -7,7 +7,7 @@
   "generalInformation": {
     "title": "Title Example of UIST Object for ABAP File Format",
     "sortPriority": 1.001,
-    "baseSpaceTemplateName": "BASE_SPACE"
+    "mergeId": "Z_MERGE_1"
```
schneidermic0, line 10: Maybe I was too fast with my approval. Is ID referring to another object name. I guess it is a reference to another space template, isn't it? 

Usually, other file formats use the term "...Name" for object names.

schneidermic0: Thanks for clarification. I guess I got confused, because it replaced a space template name and the Id in the example looked like an object name.

schneidermic0 (approved):
Looks good to me

PR #572, opened 2023-11-24
[SIAD]Add file format

aron-leibfried:


file-formats/siad/type/zif_aff_siad_v1.intf.abap
```diff
@@ -0,0 +1,43 @@
...
+
+      "! <p class="shorttext">Business Role Template ID</p>
+      "! Business Role Template ID
+      "! $required
+      BusinessRoleTemplateID   TYPE aps_iam_brt_id,
+
+      "! <p class="shorttext">Launchpad Space Template ID</p>
+      "! Launchpad Space Template ID
+      "! $required
+      LaunchpadSpaceTemplateID TYPE /ui2/space_entity_id,
```
albertmink, line 18:
```suggestion
      launchpadSpaceTemplateID TYPE /ui2/space_entity_id,
```

file-formats/siad/type/zif_aff_siad_v1.intf.abap
```diff
@@ -0,0 +1,43 @@
...
+  TYPES:
+
+    "! <p class="shorttext">Assignment</p>
+    "! Assignment
+    BEGIN OF ty_assignment,
+
+      "! <p class="shorttext">Business Role Template ID</p>
+      "! Business Role Template ID
+      "! $required
+      BusinessRoleTemplateID   TYPE aps_iam_brt_id,
```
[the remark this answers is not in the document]
albertmink: we prefer primitive data types (towards self contained, assuming basic types as `char` are known), so `aps_iam_brt_id` is a `char[30]` right?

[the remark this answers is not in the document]
albertmink: same for `/UI2/SPACE_ENTITY_ID`

file-formats/siad/type/zif_aff_siad_v1.intf.abap
```diff
@@ -0,0 +1,43 @@
...
+  PUBLIC.
+
+  TYPES:
+
+    "! <p class="shorttext">Assignment</p>
+    "! Assignment
+    BEGIN OF ty_assignment,
+
+      "! <p class="shorttext">Business Role Template ID</p>
+      "! Business Role Template ID
```
albertmink, line 11: Descriptions are sentence style (unless they are very important words/product names https://www.sapterm.com/)
```suggestion
      "! Business role template ID
```

file-formats/siad/type/zif_aff_siad_v1.intf.abap
```diff
@@ -0,0 +1,43 @@
...
+      "! $required
+      BusinessRoleTemplateID   TYPE aps_iam_brt_id,
+
+      "! <p class="shorttext">Launchpad Space Template ID</p>
+      "! Launchpad Space Template ID
```
albertmink, line 16: Sentence style

albertmink (commented):
Some minor casing issues to start with.

file-formats/siad/type/zif_aff_siad_v1.intf.abap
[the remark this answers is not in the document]
schneidermic0: I even guess you want to use underscores this will transform it to camel case in the JSON schema. Therefore, I suggest to use `business_role_template_id` or even `business_role_template`.

Since it is an object name with the length of 30 characters, you might want to reuse `zif_aff_types_v1=>ty_object_name_30`

```suggestion
      business_role_template_id   TYPE zif_aff_types_v1=>ty_object_name_30,
```

file-formats/siad/type/zif_aff_siad_v1.intf.abap
```diff
@@ -0,0 +1,43 @@
...
+  TYPES:
+
+    "! <p class="shorttext">Assignment</p>
+    "! Assignment
+    BEGIN OF ty_assignment,
+
+      "! <p class="shorttext">Business Role Template ID</p>
+      "! Business role template ID
+      "! $required
+      business_role_template_id   TYPE c LENGTH 30,
```
schneidermic0, line 13: Is fine for me. Just two minor remarks: 

1. In most cases, we call it object **names** (instead of their ID). Maybe, we can even skip it at all (I guess this would be my preferred solution) However, I would take this to the UX review what they think.
2. As mentioned before, for object names with the length of 30 we already provide a generic type which you can reuse: `zif_aff_types_v1=>ty_object_name_30` (or `if_aff_types_v1=>ty_object_name_30` within SAP systems)

file-formats/siad/type/zif_aff_siad_v1.intf.abap
```diff
@@ -0,0 +1,43 @@
...
+
+      "! <p class="shorttext">Header</p>
+      "! Header
+      "! $required
+      header         TYPE zif_aff_types_v1=>ty_header_60_cloud,
+
+      "! <p class="shorttext">General</p>
+      "! General
+      "! $required
+      assignment     TYPE ty_assignment,
```
schneidermic0, line 39: What do you want to emphasise. Do developers specify "general information" or the "assignments" of the object?

schneidermic0 (commented):
Thanks for the update 👍 

Just one more question and one more minor comment. Otherwise it looks good to me.

file-formats/siad/type/zif_aff_siad_v1.intf.abap
aron-leibfried: We compared our editor with other editors of us and we always have this "General" at top in other editors.
But you are right developers specify more the "assignments" of the object, so "Assignment" would be better?

schneidermic0: Yes, "General Information" is used, if we are not able to find a better semantic name. If there is a more specific name, we suggest to use this one. See second-last paragraph of https://github.com/SAP/abap-file-formats/blob/main/docs/json.md#writing-json-schema-with-abap-types

I am not sure, maybe, it's even "Assignements" because there are multiple assignments (to the business role template and to the space template)? It's not an assignment between business role template and space template, is it? What do you think?

aron-leibfried: I think general information would be the best choice, the whole object is an assignment of a business role template and a space template, so if it would be "Assignment(s)" this would be a duplicate and makes not that much sense in this position.

wurzka (changes requested):
We performed the UX review. See my comments.

wurzka (approved):
Thanks for the changes. Looks good to me!

PR #574, opened 2023-12-06
Make generated JSON Schema available in PR

albertmink:
Just paste all generated schema into the log of the GitHub Action. Could be improved by pasting only the schema of the specific object type, that is contained in PR.

### DIFF
--- a/.github/workflows/generate.yml
+++ b/.github/workflows/generate.yml
@@ -19,3 +19,6 @@ jobs:
       run: |
         cd generate
         npm run generate
+    - name: Output generate JSON Schema
+      run: |
+        find generate/generated -name '*.json' -print -exec cat {} \; -exec echo \;


albertmink:
New group 'Output generate JSON Schema' 
Screenshot 2023-12-06 at 13 50 56

.github/workflows/generate.yml
```diff
@@ -21,4 +21,5 @@ jobs:
         npm run generate
     - name: Output generate JSON Schema
```
larshp, line 22:
```suggestion
    - name: Output generated JSON Schema
```

larshp:
alternatively/supplement by adding it as an artifact to the run

https://github.com/actions/upload-artifact

wurzka (approved):

PR #576, opened 2023-12-15
[SCP1]Remove "Software Component" related fields

goreraks:
Initially we decided to keep "Software Component", "Minimum Release" and "Maximum Release" fields that can be maintained for BC Sets. Because of Software Components are being merged, the code split from S/4 and the issues raised while activating a BC Sets in Steampunk environment, we decided not to keep these fields in AFF.

@wurzka had requested to upload a simple example with lesser in file size. So I will create a seperate PR for the same soon.

### DIFF
--- a/file-formats/scp1/scp1-v1.json
+++ b/file-formats/scp1/scp1-v1.json
@@ -77,24 +77,6 @@
             "A hierarchical BC Set comprises several other BC Sets, which can also be hierarchical. The hierarchy can have any number of levels"
           ],
           "default": "simple"
-        },
-        "softwareComponent": {
-          "title": "Software Component",
-          "description": "BC Set is assigned to a software component. If the software component is not installed in the system, then activation of BC Set is aborted",
-          "type": "string",
-          "maxLength": 30
-        },
-        "minimumRelease": {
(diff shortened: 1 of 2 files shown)


albertmink:
This thing with the example is, the example is no longer valid with the changes in the JSON Schema. As a workaround, you could remove the data in the existing example, `softwareComponent`, `minimumRelease` and `maximumRelease`. Or you add the simplified example to this PR.

schneidermic0:
> wurzka had requested to upload a simple example with lesser in file size. So I will create a separate PR for the same soon.

I agree with Albert. We should also update the example so that it is at least is compatible with your changes.

schneidermic0:
Technically, your changes are incompatible. As far as I know SCP1 file format has not been available, yet

goreraks:
Yes, SCP1 file format is not yet available.

schneidermic0:
@goreraks There hasn't been any update to this PR for 6 weeks now. When do you plan to fix the example? Without fixing the example we cannot merge this PR in our main branch because we will see failing tests.

schneidermic0:
No update since more than two months. I have converted this to "draft"

PR #577, opened 2023-12-15
[SAJC] Add new parameter attributes: Title, Screen element, Radio button group

heichmann:
A new type of class-based application jobs is introduced. Here, the title of the parameter, the screen element which is used for the parameter and the radio button group are not set in the class any more, but in the application job catalog entry. Therefore, the application job catalog entry has to be extended.


heichmann:
Somehow, the 'File changed' section also contains all changes of the commit 'SAJC: Add Parameters, Groups, Sections' which I have done 6 monthes ago. Did I miss something in this commit? Or, is the reason that we didn't release the transport with these changes yet?

wurzka:
You created this pull request out of a branch from your fork that is not at the state of the main branch of our repository.
The best/easiest solution would be to sync the main branch of your fork, create a new branch and then push the changes to this branch. Out of that, then create a pull request.
This one here, we would close then.

PR #578, opened 2023-12-15
[SAJT] Increase size of parameter name

heichmann:
A new type of class-based application jobs is introduced. Here, the size of the parameter name is increased from 8 to 38 characters. Because [SAJT] should be able to handle old and new class-based application job templates, the size of the parameter name has to be increased.


file-formats/sajt/type/zif_aff_sajt_v1.intf.abap
```diff
@@ -57,7 +57,7 @@ INTERFACE zif_aff_sajt_v1
       "! <p class="shorttext">Name</p>
       "! Name of the parameter
       "! $required
-      name  TYPE  c LENGTH 8,
+      name  TYPE  c LENGTH 38,
```
wurzka, line 60: This could be an incompatible change if there are systems that cannot store this value. Did you think about this?

schneidermic0: How does a parameter class-based application jobs is defined? Was it extended by 30 characters, because the corresponding classname is added to the parameter name?

file-formats/sajt/README.md
```diff
@@ -1,5 +1,5 @@
 # SAJT File Format
 
 File | Cardinality | Definition | Schema | Example
-:--- | :---  | :--- | :--- | :---
-`<name>.sajt.json` | 1 | [`zif_aff_sajt_v1.intf.abap`](./type/zif_aff_sajt_v1.intf.abap) | [`sajt-v1.json`](./sajt-v1.json) | [`z_aff_example_sajt.sajt.json`](./examples/z_aff_example_sajt.sajt.json)
+:--- | :--- | :--- | :--- | :---
+`<name>.sajt.json` | 1 |  [`zif_aff_sajt_v1.intf.abap`](./type/zif_aff_sajt_v1.intf.abap)  | [`sajt-v1.json`](./sajt-v1.json) | [z_aff_example_sajt.sajt.json](./examples/z_aff_example_sajt.sajt.json)
```
schneidermic0, line 5: Did you generate the output of the readme file using our program? 
If so, there seems to be an issue with the generator for the readme file.

```suggestion
`<name>.sajt.json` | 1 | [`zif_aff_sajt_v1.intf.abap`](./type/zif_aff_sajt_v1.intf.abap) | [`sajt-v1.json`](./sajt-v1.json) | [`z_aff_example_sajt.sajt.json](./examples/z_aff_example_sajt.sajt.json)
```

heichmann: Yes, the file was generated by report SAFF_GENERATE_REPO

file-formats/sajt/type/zif_aff_sajt_v1.intf.abap
schneidermic0: OK, I understand. Then I think we can do the field extension. In lower releases you might get syntax errors if you import an object which uses functionality of higher releases. Same as for ABAP code using new statements not supported in lower releases.

file-formats/sajt/README.md
schneidermic0: I suggest to fix it manually in this PR by committing my suggestion above. We need to fix the generator.

albertmink: opened issue https://github.com/SAP/abap-file-formats-tools/issues/299. Still need to be fixed manually here.

schneidermic0: I have created also an issue to fix the generator SAP/abap-file-formats-tools#300

heichmann: I changed the file

PR #581, opened 2024-01-12
Add documentation/example for translation

albertmink:


docs/properties.md
```diff
@@ -0,0 +1,46 @@
...
+key=value
+...
+```
+The `key` refers to the text that is translated (source language), whereas `value` is the actual translated text.
+A key follows the convenction of [JSONPath](https://goessner.net/articles/JsonPath/) when refering to data in JSON files.
+
+### File Name
+
+The translation of an ABAP object are stored in file named
+`<object_name>.<object_type>.i18n.<language>.properties`
```
schneidermic0, line 15: I am not sure. Do we need the extension `.i18n`? Wouldn't the addition `.<language>` be sufficient? 

Most probably, having the extension is more "save" because it introduces a kind of key for translations.

albertmink: Just adopted what is already in AbapGit and in addition it follows what we did for textpools `<object_name>.<object_type>.texts.<language>.properties`
So this was a natural choice.

schneidermic0 (approved):
Looks good to me

PR #583, opened 2024-01-23
[UIPG] Update UIPG to support UIAD Assignment

HelgeCD:
A bit unsure on what to do with the v.1 files. For now I kept them.


schneidermic0:
ABAP file formats for UIPG has not been available in any delivered ABAP version, yet. Is my understanding correct? 

If so, I think you can just update version v1 and no v2 is needed for this update. Maybe, we can reduce complexity here. ;)
If you are not sure we can have a look at it together.

HelgeCD:
> ABAP file formats for UIPG has not been available in any delivered ABAP version, yet. Is my understanding correct?
> 
> If so, I think you can just update version v1 and no v2 is needed for this update. Maybe, we can reduce complexity here. ;) If you are not sure we can have a look at it together.

adjusted: we now only have the v1 format with the new UIAD properties :)

I am a bit unsure, why the build pipeline is now failing. Could you give details on what the issue with the schema is?

albertmink:
Looking into it

file-formats/uipg/type/zif_aff_uipg_v1.intf.abap
```diff
@@ -25,8 +25,8 @@ INTERFACE zif_aff_uipg_v1
 
   "! <p class="shorttext">Tile Display Format</p>
   "! Tile display format
-  "! $values {@link zif_aff_uipg_v1.data:co_tile_display_format}
-  "! $default {@link zif_aff_uipg_v1.data:co_tile_display_format.standard}
+  "! $values {@link /ui2/if_aff_uipg_v2.data:co_tile_display_format}
```
albertmink, line 28: Here you are referencing to data/types that are not in the repository, right? This most probably confused the test.

albertmink: Which ist ok, then the file format shall be self contained.

HelgeCD: that's what you get from not running through the entire generation process again. I forgot to adjust the the interface... unfortunately, the check is still red.

The Error Description is:
Error: UIPG: Provided and generated JSON Schema differ
Error:                   "maxLength": 2
Error:                   "maxLength": 2
What could this mean?

file-formats/uipg/type/zif_aff_uipg_v1.intf.abap
```diff
@@ -132,13 +160,21 @@ INTERFACE zif_aff_uipg_v1
       id              TYPE c LENGTH 35,
 
       "! <p class="shorttext">Type</p>
-      "! Type
+      "! type
```
schneidermic0, line 163: I still would start with a capital letter for the first word.

```suggestion
      "! Type
```

file-formats/uipg/type/zif_aff_uipg_v1.intf.abap
```diff
@@ -188,7 +224,7 @@ INTERFACE zif_aff_uipg_v1
       general_information TYPE ty_general_information,
 
       "! <p class="shorttext">Sections</p>
-      "! Sections
+      "! sections
```
schneidermic0, line 227:
```suggestion
      "! Sections
```

file-formats/uipg/type/zif_aff_uipg_v1.intf.abap
```diff
@@ -91,21 +91,45 @@ INTERFACE zif_aff_uipg_v1
...
+  TYPES:
+    "! <p class="shorttext">UIAD Assignment</p>
+    "! Assignment of a launchpad app descriptor (UIAD) item
+    BEGIN OF ty_uiad_assignment,
+
+      "! <p class="shorttext">UIAD ID</p>
+      "! ID of a launchpad app descriptor (UIAD) item
+      "! $required
+      uiad_id             TYPE ty_uiad_id,
+
```
schneidermic0, line 120: I am wondering whether users understand the term UIAD which is a transport type. 

Would it be helpful for users/developers to see something like "launchpad app descriptor" instead of "UIAD"?

schneidermic0 (commented):
Thanks for beating our checks :D 

Please see my comments. Main question from my side whether it make sense to use a different terminology instead of UIAD.

file-formats/uipg/type/zif_aff_uipg_v1.intf.abap
albertmink: The generation process knows only things that are in the repository.
`zif_aff_uipq_v1.data:co_tile_display_format` is inside the repository, `/ui2/if_aff_uipg_v2.data:co_tile_display_format` is not. Maybe stay with `zif_aff_uipq_v1.data:co_tile_display_format`? The generation process fails and yeah, the error message/description itself is not that helpfull.

file-formats/uipg/type/zif_aff_uipg_v1.intf.abap
HelgeCD: I don't like the name Launchpad App Descriptor, but from a UI perspective this is more in line how we name things. I mean in the ADT the folder for UIADs is called FLP App Descriptor Items, so it makes sense to have the same name here to be consistent.
--> renamed

file-formats/uipg/type/zif_aff_uipg_v1.intf.abap
HelgeCD: done

file-formats/uipg/type/zif_aff_uipg_v1.intf.abap
HelgeCD: done

schneidermic0 (approved):
In general, it looks good to me. 

I am still struggling a little bit with "LIAD". Maybe, in the context of UX review we come up with a better name.

wurzka (approved):
Thanks for your changes! Looks good to me.

PR #585, opened 2024-01-31
[SWCR] Update SWCR to include the description in the header

hottek:


schneidermic0 (approved):
@hottek Thanks for your contribution. Looks good to me.

PR #588, opened 2024-02-19
[UIST] - Adjust field labels + make name of UIPG mandatory

DominikvonSchoenebeck:
Some field labels needed to be adjusted (1) to be compliant to official SAP speech and (2) to be consistent in all UI places.


albertmink:
would you mind to generate the JSON Schema with the ABAP types in the PR? They seem to be out of sync.

DominikvonSchoenebeck:
JSON schema got generated based on the ABAP interface type and both got inserted in the PR. Still the workflow review fails and I can't find the reason.
Is the order important? Commit first the interface type and then the JSON schema?

albertmink:
Checks are all green, right?!

DominikvonSchoenebeck:
Oh, I see. But why there is a red X next to my committed file?

albertmink (approved):
Looks good to me.

PR #589, opened 2024-02-27
[RONT] Add ODM Entity Name and Interface Behavior Definition to RONT AFF

Ehret-Thomas:
The AFF interface for SAP Object Type (RONT) is extended by two new optional fields:
* interface_behavior_definition
* odm_entity_name

These fields make it possible to relate an existing SAP Object Type to a One Domain Model (ODM) entity and / or [Interface Behavior Definition](https://help.sap.com/doc/abapdocu_latest_index_htm/latest/en-US/abenbdl_rap_bo_interface.htm?file=abenbdl_rap_bo_interface.htm), thus enriching the SAP Object Type's integration possibilities.

### DIFF
--- a/file-formats/ront/type/zif_aff_ront_v1.intf.abap
+++ b/file-formats/ront/type/zif_aff_ront_v1.intf.abap
@@ -17,10 +17,10 @@ INTERFACE zif_aff_ront_v1
       configuration_object TYPE c LENGTH 2 VALUE `CO`,
       "! <p class="shorttext">Dependent Object</p>
       "! Dependent object
-      dependent_object TYPE c LENGTH 2 VALUE `DO`,
+      dependent_object     TYPE c LENGTH 2 VALUE `DO`,
       "! <p class="shorttext">Hierarchy Object</p>
       "! Hierarchy object
-      hierarchy_object TYPE c LENGTH 2 VALUE `HO`,
+      hierarchy_object     TYPE c LENGTH 2 VALUE `HO`,
     END OF co_category.
 
   TYPES:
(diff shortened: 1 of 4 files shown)


file-formats/ront/type/zif_aff_ront_v1.intf.abap
```diff
@@ -17,39 +17,47 @@ INTERFACE zif_aff_ront_v1
...
+      name                          TYPE c LENGTH 30,
 
       "! <p class="shorttext">Object Type Code</p>
       "! The object type code uniquely identifies the SAP Object Type.
-      object_type_code TYPE c LENGTH 5,
+      object_type_code              TYPE c LENGTH 5,
+
+      "! <p class="shorttext">Interface Behavior Definition</p>
+      "! The interface behavior definition related to the SAP Object Type.
+      interface_behavior_definition TYPE c LENGTH 30,
```
wurzka, line 56: We struggled with the term Interface Behavior Definition. Is it a special interface or a special behavior definition? For us, it sounds very technical. But we may need more input on this field.

file-formats/ront/type/zif_aff_ront_v1.intf.abap
```diff
@@ -17,39 +17,47 @@ INTERFACE zif_aff_ront_v1
...
+
+      "! <p class="shorttext">ODM Entity Name</p>
+      "! The One Domain Model entity related to the SAP Object Type.
+      odm_entity_name               TYPE c LENGTH 255,
```
wurzka, line 60: Is ODM a common abbreviation or would it be better to use One Domain Model for the title and component name?
We also recommend not to use `name` therefore our suggestion would be
```suggestion
      "! <p class="shorttext">One Domain Model Entity</p>
      "! The One Domain Model entity related to the SAP Object Type.
      one_domain_model_entity               TYPE c LENGTH 255,
```
or alternatively with the abbreviation:
```suggestion
      "! <p class="shorttext">ODM Entity</p>
      "! The One Domain Model entity related to the SAP Object Type.
      odm_entity               TYPE c LENGTH 255,
```

wurzka (changes requested):
Hi @Ehret-Thomas, thanks for updating the file format. We had a look from UX point of view. Please see my comment.

schneidermic0:
Usually, we don't recommend to put fields on top-level anymore. However, since this would require an increase of the format version I think we should keep the structure as suggested by @Ehret-Thomas

albertmink:
Checked the terms and their context. This are kind of standard expressions/names in RAP, resp. DOM context.

albertmink (approved):
LGTM

PR #594, opened 2024-03-18
[SPRV] AFF field length extended to match DB fields

delucapietro:
The field 'namespace' was displayed with only 30 characters while it could be 255 characters long on DB.
In the wizard this namespace can be entered with more characters only when displaying an SPRV in ADT the length is shortened.


wurzka:
@schneidermic0 could you please double-check? From my point of view, this change should not request an update of the format version.

schneidermic0:
> @schneidermic0 could you please double-check? From my point of view, this change should not request an update of the format version.

https://github.com/SAP/abap-file-formats/blob/main/docs/json.md#incompatible-file-format-changes states, it is an incompatible change if:

> Length of a field is extended if some systems cannot store this information

I understand @delucapietro's PR description that the DB can store the length of 255 characters in all releases. 

> The field 'namespace' was displayed with only 30 characters while it could be 255 characters long on DB.

If so, I would consider it as a compatible change.

schneidermic0:
The change looks good to me.

@delucapietro Is my assumption above correct?

delucapietro:
@schneidermic0 yes the assumption is correct DB was always in all releases capable of storing the data with 255 characters only the display was shortened by mistake.

schneidermic0 (approved):
@delucapietro Thanks for the update. Looks good to me

PR #600, opened 2024-03-21
Minka/wip

albertmink:


compatibility-check/json-schema-diff.js
```diff
@@ -0,0 +1,61 @@
...
+  for (const schema of changedSchema) {
+
+    const dataNew = readFileSync(`../${schema}`, 'utf8');
+    const schemaNew = JSON.parse(dataNew);
+
+    try {
+      await exec.exec(`git checkout remotes/origin/main -- `, [schema], { cwd: `../`} );
+    } catch (error) {
+      core.info(`File ${schema} is not known to main branch.`);
+      // file is not on main branch, so we continue and compare the file to itself (no harm)
```
schneidermic0, line 41: Can't we even return? Nothing needs to be check, right?

albertmink, line 41:
```suggestion
      continue;
```

schneidermic0 (approved):
Thanks, @albertmink. Looks good to me.

PR #601, opened 2024-03-22
SITO Version 2

micotto:
Hi colleagues,
due to incompatible changes, the version for object SITO needs to be updated.
The change consists of adding one new mandatory field (personal_data)
and one new entity (personal_data_fields).
(Backend implementation is not done, so I could not create a new example object yet.)
Best regards,
Michael


file-formats/sito/type/zif_aff_sito_v2.intf.abap
```diff
@@ -280,16 +314,16 @@ INTERFACE zif_aff_sito_v1
...
       "! $required
-      name        TYPE c LENGTH 40,
-      "! <p class="shorttext">Description</p>
-      "! Description
+      name            TYPE c LENGTH 40,
+      "! <p class="shorttext">Long Description</p>
+      "! Long description
       "! $required
-      description TYPE c LENGTH 255,
+      longdescription TYPE c LENGTH 255,
```
schneidermic0, line 326: Do you want to rename this? I saw (in the old example) there are other "description" fields in text sections.

See: https://github.com/SAP/abap-file-formats/blob/23986fa71df0115e9d7fe9325ffe7bc2de986211/file-formats/sito/examples/zmo_sito_test.sito.json#L25

schneidermic0 (changes requested):
Thanks, @micotto for updating the file format.

In general I think it looks good. I struggle with the rename of the description field (see below).

In general, we should keep also version 1 in the repository. This was done also for other object types. E.g., see https://github.com/SAP/abap-file-formats/tree/main/file-formats/cfdf

However, adding the incompatible change to the existing (renamed) files first was helpful for me reviewing the changes. So I could easily identify what you have changed :)

schneidermic0:
Just a remark: There are a lot of commits in this PR. It seems to me you were using your old branch of v1. I guess I would have created a new branch (based on latest main branch) instead of reusing the existing branch of your fork.

albertmink:
> Just a remark: There are a lot of commits in this PR. It seems to me you were using your old branch of v1. I guess I would have created a new branch (based on latest main branch) instead of reusing the existing branch of your fork.

Looks like there is a lot of renaming of files happening. So e.g. `abc-v1.json` becomes `abc-v2.json`, which deletes basically the `abc-v1.json`. Instead, you want to add a new `abc-v1.json` 😃

micotto:
Hi @schneidermic0 , thanks for your quick reaction.

Regarding my use of the old branch with many commits, does it still make sense to create a new one ? I would just do it the way you prefer it. 

For the description field I will need to discuss internally, or maybe you also have a good idea. Our fundamental issue is the following: A Situation Object contains fields "id", "name", "description". Mapping to ADT notation, "id" corresponds to the name of the artifact, whereas "name" is what we want to display in the headers' (and in the properties section's) "description" field. Moreover, the Situation Object's "description" field contains a long description. It is currently located at sub-entity "text". 
If, in our AFF, we replace "longdescription" with "description" we fear there might be confusion with the differing values in header's and properties section's description.

I'll take care of uploading both versions. (This was not clear for me from your Wikis (which I like a lot :-) ). Maybe somewhere in the "Changing ABAP File Formats" section of your "Generic Editor FAQ" you could describe expected upload structure, similar to the specification in section "What the pull request should include" of "Add object to github repository".)

schneidermic0:
> Regarding my use of the old branch with many commits, does it still make sense to create a new one ? I would just do it the way you prefer it.

No, keep this PR as it is. Otherwise we lose our conversation. ;)

schneidermic0:
> For the description field I will need to discuss internally, or maybe you also have a good idea. 

I am afraid, I have also no good idea, how to handle this. The tricky thing starts IMHO to have multiple descriptions. I am even wondering whether this should be real documentation. If so a dedicated KTD documentation might be a solution.

In any case, it should be at least consistent within the one object type. Now, we have fields called "longdescription" and "description" for (as far as I understand) on the same level of descriptions (in section "texts").

micotto:
Hi. With the latest push, files referring to version 1 are included now.
Still need to discuss the naming of description fields internally (when colleagues are back from Easter vacation).
One question: 
The new field personal_data is defined as required and equipped with a list of enumeration values. When I create a Situation Object now (based on version 2), the new configuration seems to have no effect (the field does not show when I press "STRG-Space"; it is not bold-faced when I write it myself; possible values do not show when I press "STRG-Space"). Not sure if I am missing something in my configuration, or if version 2 needs to be activated somehow from your end ?

schneidermic0:
> One question:
The new field personal_data is defined as required and equipped with a list of enumeration values. When I create a Situation Object now (based on version 2), the new configuration seems to have no effect (the field does not show when I press "STRG-Space"; it is not bold-faced when I write it myself; possible values do not show when I press "STRG-Space"). Not sure if I am missing something in my configuration, or if version 2 needs to be activated somehow from your end ?

I have just dropped you a mail since the question is not related to the definition of the ABAP file formats. ;)

schneidermic0:
> Still need to discuss the naming of description fields internally (when colleagues are back from Easter vacation).

You might want to consider to move this pull request to "draft" as long as you are working on this

micotto:
> > For the description field I will need to discuss internally, or maybe you also have a good idea.
> 
> I am afraid, I have also no good idea, how to handle this. The tricky thing starts IMHO to have multiple descriptions. I am even wondering whether this should be real documentation. If so a dedicated KTD documentation might be a solution.
> 
> In any case, it should be at least consistent within the one object type. Now, we have fields called "longdescription" and "description" for (as far as I understand) on the same level of descriptions (in section "texts").

After internal discussions our plan is now the following:
-rename field "longdescription" to "description"
-remove field "name" from entity situation_object-text (as described earlier, this field contains redundant information)
I would like to hear your opinion especially on the removal of "name" before I start adapting the backend.
Thanks.

schneidermic0:
> -remove field "name" from entity situation_object-text (as described earlier, this field contains redundant information)
I would like to hear your opinion especially on the removal of "name" before I start adapting the backend.

In general, removing fields is incompatible. But, since you anyhow update your format version and if "name" contains redundant information, I would remove it, too.

> -rename field "longdescription" to "description"
At least this is consistent with all the other places. I am not sure how this particular field interferes with the field `header-description`. Is it also redundant?

micotto:
> > -remove field "name" from entity situation_object-text (as described earlier, this field contains redundant information)
> > I would like to hear your opinion especially on the removal of "name" before I start adapting the backend.
> 
> In general, removing fields is incompatible. But, since you anyhow update your format version and if "name" contains redundant information, I would remove it, too.
> 
> > -rename field "longdescription" to "description"
> > At least this is consistent with all the other places. I am not sure how this particular field interferes with the field `header-description`. Is it also redundant?

Field "longdescription" contains information that is not redundant. ("header-description" contains exactly the same as "situation_object-text-name".)

schneidermic0:
OK, I see. Seems to be confusing

schneidermic0:
> The fields containing descriptions are now adapted. A version 2 example is included as well.
From our point of view, version 2 is complete.

I undsterstand you are ready from your side :). I suggest you fix the end of file issue in the example and change this PR to "Ready for review"

micotto:
Hi colleagues. As requested, EoF issue is resolved. We are ready for review now.

schneidermic0 (commented):
Looks good to me now. 

One minor question: Do you think we can use the same example object for version 1 and 2? I think this would help to see the differences between both versions

micotto:
I replaced the version 1 example with the predecessor of the version 2 example. Is this what you had in mind ? (I agree that's better for comparison.)
There is a new error showing up after my push. Not sure if this is something to be fixed on our end ?

schneidermic0:
> There is a new error showing up after my push. Not sure if this is something to be fixed on our end ?

I have updated the readme file. The link to example v2 was missing and v1 was pointing to the wrong/old filename. The latter was the reason for the error reported by the checks

schneidermic0 (approved):
Looks good to me, now.

albertmink (approved):
LGTM

schneidermic0:
> Hi colleagues. It seems that version 2 is complete (from your perspective as well).

I have just waited whether others want to review, too.

> Thank you for the once again fast review process and for all your help.
:-)

Happy to hear. Thank you for your contributions.

PR #604, opened 2024-03-25
Generate JSON Schema for formatVersion != 1 

### LINKED ISSUE #603: Generate JSON Schema fails for version != 1
To me this looks like if the generation always looks for a type with `V1`.
https://github.com/SAP/abap-file-formats/blob/23986fa71df0115e9d7fe9325ffe7bc2de986211/generate/cl_run.clas.abap#L22
See
image

This might cause the errors in #601 .

We didn't found this earlier since there was always a V1. In this PR there is accidentally no V1, so this fails.

Other object type with V2 is [cfdf](https://github.com/SAP/abap-file-formats/tree/main/file-formats/cfdf)

albertmink:
closes #603 

Console log/output of the GitHub Action
Before
Screenshot 2024-03-26 at 09 48 58

After
Screenshot 2024-03-26 at 09 48 35

### DIFF
--- a/generate/aff.mjs
+++ b/generate/aff.mjs
@@ -73,6 +73,11 @@ async function run() {
           fs.writeFileSync(filename, result.get());
 
 
+          if (!fs.existsSync(filename_aff)){
+            core.warning('Failed to access file: '+filename_aff);
+            continue;
+          }
+
           const command = `diff ${filename_aff} ${filename}`;
           const output = child_process.execSync(`${command} || true`);
           if (output.toString().length > 0) {
--- a/generate/aff.mjs
+++ b/generate/aff.mjs
@@ -69,7 +69,7 @@ async function run() {
 
           const result = await abap.Classes["CL_RUN"].run({ object_type: new abap.types.String().set(aff.object_type), format_version: aff.format_version });
           const filename = `generated` + path.sep + aff.object_type + `-v`+aff.format_version+`.json`;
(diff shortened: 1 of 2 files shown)


albertmink:
Almost there. Sub-types as FUNC, REPS are not covered yet, also without warning 😬 

Screenshot 2024-03-26 at 09 01 36

albertmink:
Need logic from here (main-object//sub-object)
https://github.com/SAP/abap-file-formats-tools/blob/609a2a390577944cbd9ee630751a6df38b12283c/src/z_generate_repo.prog.abap#L674-L681

generate/aff.mjs
```diff
@@ -31,38 +31,70 @@ async function run() {
...
-      createAnnotations(output.toString(), path.resolve(filename_aff));
-      //core.info(command);
-      //core.info(output.toString());
-    } else {
-      core.notice(type + " success");
+    const files = fs.readdirSync(directory + path.sep + type);
+    const jsonFiles = files.filter(file => path.extname(file) === '.json');
+    let objTypeVersNumb = jsonFiles.map((file) => {
+      // Extract the characters before '-v' as namePart and the number after '-v' as versionNumber using regex
+      let match = file.match(/(.*)(?:-v)(\d+)(?=\.json)/);
```
schneidermic0, line 54: Later, when the functionality might generate the schema, I guess we cannot use this approach to identify the version number (because the file might not exist, yet). However, for now it seems to be fine to me.

schneidermic0 (approved):
Looks good to me

PR #610, opened 2024-04-11
Consider modified JSON Schema only

albertmink:
Sort out the non-relevant JSON Schema right at the beginning, so we don't need to catch those kind of failures later. E.g. line 27-28 will be removed by this change 
image


schneidermic0 (approved):
Looks good to me. Thanks :)

PR #612, opened 2024-05-03
[AIFD] Add new object type AIFD

D047539:


file-formats/aifd/README.md
schneidermic0, line 1: Example is missing, as of now

file-formats/aifd/type/zif_aff_aifd_v1.intf.abap
```diff
@@ -0,0 +1,41 @@
...
+
+    "! <p class="shorttext">AIFD Object Type</p>
+    "! Object type AIFD
+    BEGIN OF ty_main,
+      "! $required
+      format_version TYPE zif_aff_types_v1=>ty_format_version,
+      "! <p class="shorttext">Header</p>
+      "! Header
+      "! $required
+      header         TYPE zif_aff_types_v1=>ty_header_60_src,
```
schneidermic0, line 35: Most probably object type `..._src` is wrong in this case.

Which ABAP language versions do you plan to support? Cloud development and key user?

file-formats/aifd/type/zif_aff_aifd_v1.intf.abap
```diff
@@ -0,0 +1,41 @@
...
+      "! <p class="shorttext">Namespace</p>
+      "! Namespace
+      namespace         TYPE c LENGTH 6,
+      "! <p class="shorttext">Interface Name</p>
+      "! Interface Name
+      interface_name    TYPE c LENGTH 10,
+      "! <p class="shorttext">Interface Version</p>
+      "! Interface Version
+      interface_version TYPE c LENGTH 5,
+    END OF ty_aif_interface,
```
schneidermic0, line 21: Sorry, I am not an expert in AIF. What is the difference between an interface object name and an interface name?

schneidermic0 (commented):
See my first comments/questions below.

Do you think it is possible to provide an example? I guess this will help us to review the object type.

file-formats/aifd/README.md
D047539: How can I create the example?

file-formats/aifd/type/zif_aff_aifd_v1.intf.abap
D047539: type ..._src replaced by if_aff_types_v1=>ty_header_60

For ABAP language version we only support Standard and Cloud development.

file-formats/aifd/type/zif_aff_aifd_v1.intf.abap
D047539: The interface object name is the technical key in the TADIR for the AIFI (AIF Interface) development object.
The AIF Interfaces consists of Namesapce, Interface Name and Interface Version.
Example:
      "objectName": "/AIFTCUST_WS_I00001",
      "namespace": "/AIFT",
      "interfaceName": "CUST_WS_I",
      "interfaceVersion": "00001"

file-formats/aifd/type/zif_aff_aifd_v1.intf.abap
schneidermic0: Sorry for the late response. If you "only" support `standard` and `cloudDevelopment`, I suggest to use  `zif_aff_types_v1=>ty_header_60_cloud`

file-formats/aifd/examples/z_aff_example_aifd.aifd.json
```diff
@@ -0,0 +1,22 @@
...
+    "description": "Example AIFD for ABAP File Format",
+    "originalLanguage": "en",
+    "abapLanguageVersion": "cloudDevelopment"
+  },
+  "aifInterfaces":[
+    {
+      "objectName":"/AIFTCUST_WS_I00001",
+      "namespace":"/AIFT",
+      "interfaceName":"CUST_WS_I",
+      "interfaceVersion":"00001"
```
schneidermic0, line 13: Isn't the data redundant? So wouldn't it be sufficient to either have only the `objectName` or the fields `namespace`, `interfaceName`, `interfaceVersion`?

schneidermic0 (commented):
I am sorry for the late response. I did not manage to look into it earlier.

Thanks for providing the example. This was helpful for me. 👍 

I try to first have a look at the general structure. Most probably I suggest to use different names, later as soon as you think the structure is fine.

file-formats/aifd/examples/z_aff_example_aifd.aifd.json
D047539: No, the data is not necessarily redundant. It may be that the interface object name is derived from the interface keys namespace, interface name and interface version, but this is not mandatory and can be different if necessary.
And in exactly such a case, when the interface object name is different, an additional display of the interface keys makes sense.

D047539: What do you mean with use different names?

file-formats/aifd/type/zif_aff_aifd_v1.intf.abap
D047539: type replaced by if_aff_types_v1=>ty_header_60_cloud as suggested

file-formats/aifd/examples/z_aff_example_aifd.aifd.json
schneidermic0: OK, I see. 👍 

> What do you mean with use different names?

Maybe, we suggest different field names. But we have to double-check this aspect first.

file-formats/aifd/type/zif_aff_aifd_v1.intf.abap
```diff
@@ -0,0 +1,40 @@
+INTERFACE zif_aff_aifd_v1
+  PUBLIC.
+  TYPES:
+    "! <p class="shorttext">AIF Interface</p>
+    "! AIF interface
+    BEGIN OF ty_aif_interface,
+      "! <p class="shorttext">Interface Object Name</p>
```
schneidermic0, line 7: Just a minor: 

Usually, we try to keep the title (= "shorttext") and the field name in sync. Would it make sense to decide either for "Object Name" or "Interface Object Name".

If I understood your explanation I guess I would go for just "Object Name" or even "Name" (the latter we used in various other object types)

file-formats/aifd/type/zif_aff_aifd_v1.intf.abap
```diff
@@ -0,0 +1,40 @@
...
+      "! Interface Object Name
```
schneidermic0, line 8: Please, use sentence style for the description:

```suggestion
      "! Interface object name
```

file-formats/aifd/type/zif_aff_aifd_v1.intf.abap
```diff
@@ -0,0 +1,40 @@
...
+      "! $required
+      object_name       TYPE c LENGTH 40,
+      "! <p class="shorttext">Namespace</p>
+      "! Namespace
+      namespace         TYPE c LENGTH 6,
+      "! <p class="shorttext">Interface Name</p>
+      "! Interface Name
```
schneidermic0, line 15:
```suggestion
      "! Interface name
```

file-formats/aifd/type/zif_aff_aifd_v1.intf.abap
```diff
@@ -0,0 +1,40 @@
...
+      interface_name    TYPE c LENGTH 10,
+      "! <p class="shorttext">Interface Version</p>
+      "! Interface Version
```
schneidermic0, line 18:
```suggestion
      "! Interface version
```

file-formats/aifd/examples/z_aff_example_aifd.aifd.json
```diff
@@ -0,0 +1,22 @@
...
+    },
+    {
+      "objectName":"/AIFTCUST_WS_O00001",
+      "namespace":"/AIFT",
+      "interfaceName":"CUST_WS_O",
+      "interfaceVersion":"00001"
```
schneidermic0, line 19: Just one more question: Since only `objectName` is the only required field, `namespace`, `interfaceName` and `interfaceVersion` can be empty. Is this the behavior you intend?

schneidermic0 (dismissed):
I guess I understood your object's structure, now. This is fine for me. 

I just found some minor issues/questions.

file-formats/aifd/examples/z_aff_example_aifd.aifd.json
D047539: No, the fields namespace, interfaceName and interfaceVersion should be filled, but not changeable. The fields should be filled automatically by filling the objectName. The fields can be used for the objectName search, as the value of the ObjectName may be a generated and not readable value, so the fields are needed to identify the assigned AIF interfaces.

schneidermic0 (commented):
@D047539 Thanks for all your adaptations. In general it looks good to me. 👍 

However, I am still struggling with the different interface names and object names and I am not sure whether I get your requirements. Do you think it make sense to discuss this in a short call with you, @wurzka and me?

file-formats/aifd/examples/z_aff_example_aifd.aifd.json
wurzka: I had a call with @D047539: the important field here is the `interfaceObjectName`/`objectName`. This field does not need to be derived from the other 3 fields. These fields, however, can help to better identify the chosen object. But they are not needed for the AIFD object. Hence, we decided to remove these fields and only keep the necessary field `interfaceObjectName`/`objectName`.

wurzka (approved):
Thanks for the adaption. Looks good to me, now.

file-formats/aifd/examples/z_aff_example_aifd.aifd.json
```diff
@@ -0,0 +1,19 @@
+{
+  "formatVersion": "1",
+  "header": {
+    "description": "Example AIFD for ABAP File Format",
+    "originalLanguage": "en",
+    "abapLanguageVersion": "cloudDevelopment"
+  },
+  "aifInterfaces":[
+    {
+      "interfaceObjectName":"INT_OBJ_1"
```
albertmink, line 10: @D047539 did you consider to change `interfaceObjectName` to `name`? Typically, the array name, here 'aifInterfaces', is sufficient and in particular if the array element contains a single String.

D047539: @albertmink no, we have not changed the name from interfaceObjectName to name because name is very generic one and from our point of view it is not a clear name. We would like to keep the interfaceObjectName as name.

albertmink (approved):
LGTM

PR #621, opened 2024-06-17
[UIAD] - Smooth out the structure of Custom tile and its properties

ViktoriaFreidel:
We reduce a substructure of custom tiles to be in sync with DB definition and for better Layout control
and side effects


ViktoriaFreidel:
Could you help to understand and resolve compatibility check error? We change a sub structure, should we then use   "formatVersion": "2" ? Or there is on other reason of errors? Thank you.

albertmink:
The Compatability Check helps to detect incompatible changes. It is however not 'required' and doesn't block merging.

Did you shipped the editor? If not, you might just ignore this check.

ViktoriaFreidel:
Currently we are in development phase and shipment of the editor not done yet. 
I would to ask you for review.

albertmink (approved):
Thanks for the update. LGTM

PR #632, opened 2024-07-10
Feature/aifd

D047539:
AIFD: Adaption of title interface


albertmink:
@D047539 not sure what happens, but the PR does not looks like there is some modification. It is all new, see 
Screenshot 2024-07-11 at 12 19 48
Update your fork fist, then apply changes and open PR.

PR #634, opened 2024-07-11
[AIFR] Add new object type AIFR

D047539:


huber-nicolas:
Can you please create a new pull request for AIFR which is based on the current main branch (so that we do not see the AIFD changes here)

huber-nicolas (changes requested):
Please see my comment. Can you please remove the AIFD changes from this pull request

PR #635, opened 2024-07-11
[AIFD] Change in title

D047539:


albertmink:
Requires only review from @SAP/abap-file-formats-ux

albertmink (approved):
LGTM

file-formats/aifd/type/zif_aff_aifd_v1.intf.abap
```diff
@@ -23,8 +23,8 @@ INTERFACE zif_aff_aifd_v1
       "! Header
       "! $required
       header         TYPE zif_aff_types_v1=>ty_header_60_cloud,
-      "! <p class="shorttext">AIF Interfaces</p>
-      "! AIF interfaces
+      "! <p class="shorttext">Interfaces</p>
+      "! Interfaces
       "! $required
       aif_interfaces TYPE ty_aif_interfaces,
```
wurzka, line 29: The name of the component still contains `aif`. We usually recommend having the name and the title in sync. For me, it is fine here. However, I want to make sure, that in the JSON field (for example in https://github.com/SAP/abap-file-formats/blob/main/file-formats/aifd/examples/z_aff_example_aifd.aifd.json), the component is still named `aifInterface`.
If the presence of `aif` is ok for you, we can merge this change

wurzka (commented):
Just one question, see my comment

file-formats/aifd/type/zif_aff_aifd_v1.intf.abap
D047539: Unfortunately, AIF is not an official abbreviation for our product, so it is not allowed to be seen on the UI. However, as a component name in the background it should be ok from our point of view and gives a little more detail about what is displayed and maintained.

wurzka: But that's my point. It is not only in the background. It is in the JSON file, part of the "official" AFF. So, if `aif` should not be seen, I suggest removing it at least from the component name.

D047539: I need to align with our UA developer and PO how to proceed with the abbreviation AIF

D047539: We will stay with abbreviation AIF

albertmink:
No changes -> closing the issue

PR #639, opened 2024-07-22
Ensure that examples don't get deleted

albertmink:
This GitHub Action fails if there are examples deleted in pull requests. Sometimes this happens when increasing format version. For the moment it might be sufficient if this Action isn't required, the red cross typically is frightening already enough.


wurzka (approved):
Thanks, looks good to me.

Markus1812 (approved):
Looks good 👌

huber-nicolas (approved):

PR #643, opened 2024-07-29
DCAT : Added default value to sensitivity_level field.

C5314720:


file-formats/dcat/dcat-v1.json
```diff
@@ -77,7 +77,8 @@
           "enumDescriptions": [
             "Sensitive personal data",
             "Personal data"
```
wurzka, line 79: This is only cosmetic: Did you think of having the default value as first value? 
```suggestion
             "Personal data",
            "Sensitive personal data"
```
This would also mean switching the component in the constant but I do not expect any problems with this change.

wurzka (commented):
Hi @C5314720,
thanks for providing the update. It looks good to me. However, I have one small cosmetic question for that I would like to have your feedback.

file-formats/dcat/dcat-v1.json
C5314720: Change the sequence as suggested.

wurzka (approved):
Thank you for updating the file format. Looks good to me!

PR #644, opened 2024-07-30
Add recommendation for `$required` regarding boolean fields

### LINKED ISSUE #642: Add recommendation for `$required` in boolean fields in documentation
It happens from time to time that contributors set boolean fields to required. 

For boolean fields, we usually recommend to remove the `$required` annotation, because not setting it is equivalent with `false`. Setting it to required might indicate that you have to flag it. You might know this in forms on websites to accept the web site's data privacy statement.

If the field should be always visible in the JSON, `$showalways` could be used.

# Todos

- [ ] Update documentation for annotation `$required`
https://github.com/SAP/abap-file-formats/blob/main/docs/json.md#required-fields

Markus1812:
resolves #642

### DIFF
--- a/docs/json.md
+++ b/docs/json.md
@@ -154,7 +154,7 @@ If a field is to be declared as "required" in the JSON Schema, the annotation
 ```
 is used.
 
-Please note, we do not recommend using the `$required` annotation for boolean fields, as not setting it is equivalent to `false`.
+Please note, we do not recommend using the `$required` annotation for boolean fields, as omitting such a field in the JSON file is equivalent to setting its value to `false`.
 If the field is crucial to be serialized (even if the value is `false`), you can use the annotation `$showAlways`.
 
 ### Always Shown Fields
--- a/docs/json.md
+++ b/docs/json.md
@@ -154,6 +154,9 @@ If a field is to be declared as "required" in the JSON Schema, the annotation
 ```
 is used.
 
+Please note, we do not recommend using the `$required` flag for boolean fields, as not setting it is equivalent to `false`. 
+If you want the field to always be visible in the JSON, you can use the annotation `$showAlways`. 
+
 ### Always Shown Fields
 Normally, if an ABAP object is serialized, only the components of the corresponding type with a non-initial value are written to the JSON data file. If a component shall be shown regardless to its value, the annotation
 ```abap
--- a/docs/json.md
+++ b/docs/json.md
@@ -154,8 +154,8 @@ If a field is to be declared as "required" in the JSON Schema, the annotation
 ```
 is used.
 
-Please note, we do not recommend using the `$required` flag for boolean fields, as not setting it is equivalent to `false`. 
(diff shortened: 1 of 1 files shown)


docs/json.md
```diff
@@ -154,6 +154,9 @@ If a field is to be declared as "required" in the JSON Schema, the annotation
 ```
 is used.
 
+Please note, we do not recommend using the `$required` flag for boolean fields, as not setting it is equivalent to `false`.
```
schneidermic0: Usually, I think we call all the $...-stuff "annotation".

```suggestion
Please note, we do not recommend using the `$required` annotation for boolean fields, as not setting it is equivalent to `false`. 
```

docs/json.md
```diff
@@ -154,6 +154,9 @@ If a field is to be declared as "required" in the JSON Schema, the annotation
 ```
 is used.
 
+Please note, we do not recommend using the `$required` flag for boolean fields, as not setting it is equivalent to `false`. 
+If you want the field to always be visible in the JSON, you can use the annotation `$showAlways`.
```
schneidermic0: Maybe, the recommendation should be more like this. If the field is crucial to be serialized (even if the value is `false`), you can use the annotation `$showAlways`. What do you think?

schneidermic0 (changes requested):
Thanks, @Markus1812 for taking over. 👍 

Basically, it is fine. See my suggestions for improvements, below.

docs/json.md
Markus1812: Yes, you are right. Changed it.

docs/json.md
Markus1812: Yes, sounds better as it should not get the standard to add `$showAlways` to all boolean fields.

docs/json.md
```diff
@@ -154,6 +154,9 @@ If a field is to be declared as "required" in the JSON Schema, the annotation
 ```
 is used.
 
+Please note, we do not recommend using the `$required` annotation for boolean fields, as not setting it is equivalent to `false`.
```
wurzka, line 157: I struggled with this formulation. For me it means: We do recommend using the annotation because not setting the annotation is equivalent to `false`. 
Maybe we can have something like 
```suggestion
Please note, we do not recommend using the `$required` annotation for boolean fields, as not setting such a field is equivalent to `false`.
```
to make clear that it is about setting fields and not annotations. Feel free to reformulate my suggestion in better english 😉

schneidermic0: Do you prefer following suggestion?

```suggestion
"Please note, we do not recommend using the $required annotation for boolean fields, as omitting such a field in the JSON file is equivalent to setting its value to false."
```

wurzka: Yes, I like it

Markus1812: Yes, me too. Changed it.

schneidermic0 (approved):
LGTM, now. 👍 

Thanks, @Markus1812.

PR #649, opened 2024-08-02
[SAJC]: Add value help to parameters

heichmann:


file-formats/sajc/README.md
```diff
@@ -2,4 +2,4 @@
 
 File | Cardinality | Definition | Schema | Example
 :--- | :--- | :--- | :--- | :---
-`<name>.sajc.json` | 1 | [`zif_aff_sajc_v1.intf.abap`](./type/zif_aff_sajc_v1.intf.abap) | [`sajc-v1.json`](./sajc-v1.json) | [`z_aff_example_sajc.sajc.json`](./examples/z_aff_example_sajc.sajc.json)
+`<name>.sajc.json` | 1 | [`zif_aff_sajc_v1.intf.abap`](./type/zif_aff_sajc_v1.intf.abap) | [`sajc-v1.json`](./sajc-v1.json) | [`z_aff_example_sajc2.sajc.json`](./examples/z_aff_example_sajc2.sajc.json)
```
huber-nicolas, line 5:
```suggestion
`<name>.sajc.json` | 1 | [`zif_aff_sajc_v1.intf.abap`](./type/zif_aff_sajc_v1.intf.abap) | [`sajc-v1.json`](./sajc-v1.json) | [`z_aff_example_sajc.sajc.json`](./examples/z_aff_example_sajc.sajc.json)
```

Instead of creating a new example, you can also use the old example and set the value help and value help type at a parameter and serialize the object again(in the system). Please do not rename or add a second example object in this repository.

heichmann: I can remove one of the examples. Just two questions about this:
- We increased the size of the parameter names with one of the last merges. Should the example use short parameter names (to work with all releases), or should it use long parameter names (to represent the latest version)?
- Is it necessary that the example of the job catalog entry (SAJC) fits to the example of the job template (SAJT)? So, should they use the same parameter names?

huber-nicolas: > I can remove one of the examples. Just two questions about this:
> 
> * We increased the size of the parameter names with one of the last merges. Should the example use short parameter names (to work with all releases), or should it use long parameter names (to represent the latest version)?
The example should represent the latest version (long parameter names)
> * Is it necessary that the example of the job catalog entry (SAJC) fits to the example of the job template (SAJT)? So, should they use the same parameter names?
To better understand the connection between the two objects it would be good to have the parameter names in sync

heichmann: I removed the old example and I adapted the SAJT example

Markus1812: Thank you for updating the examples.

Markus1812 (approved):
Thank you for updating your AFF. We will follow up with a ux-review performed by our colleagues.

wurzka (approved):
Looks good to me

wurzka:
As adding non mandatory fields is compatible, there is no requirement for updating the format version. Therefore, we can merge the changes.

PR #650, opened 2024-08-06
Fix incorrect example in docs/json.md documentation

jonassobotta:
In the example I used for the pattern annotation I had a typo in the pattern; [a-Z]* is not a valid regex, therefore I changed it to [a-z]*


schneidermic0 (approved):
Thanks. Looks good to me.

PR #652, opened 2024-08-08
[SAJT] New example fitting to SAJC example

heichmann:
Because the SAJC example was replaced, the SAJT example has to be replaced, too, that it fits to it.


wurzka (approved):
Thanks for updating the example!

PR #656, opened 2024-08-20
React on UX Review for Object Type AIFR

D047539:


albertmink:
@D047539 you can ignore the `Compatibility Check` as it is not required.

albertmink (approved):
Thanks for renaming. LGTM

PR #662, opened 2024-09-25
[SUCO] Add new object type SUCO

WDFdaniel:

PR #661, opened 2024-09-17
[EDCC] Add new object type for EDCC

KUMARMUKULSAP:

PR #614, opened 2024-05-21
[SAIA] Add new object type SAIA

uenal-akkaya:

PR #665, opened 2024-10-01
Update abaplint.jsonc

huber-nicolas:
temporarily add exemption for ty_header_80 until SUCO is merged

### DIFF
--- a/abaplint.jsonc
+++ b/abaplint.jsonc
@@ -195,7 +195,7 @@
     "unsecure_fae": true,
     "unused_methods": true,
     "unused_types": {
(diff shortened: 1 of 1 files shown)


Markus1812 (approved):

albertmink:
won't help #662 as long as they don't update from current main. If they update, then they will use the type and the exemption is redundant.

Markus1812:
Will help for #661, #614 and main branch since they all had failing abaplint checks. The exemption can be removed again after #662 is merged (because `ty_header_80` is used then).

PR #672, opened 2024-11-04
[SMBC] update 2411

sepp4me:
new attributes:

- showTransportSelectionStrip
- selectAll
- frozenColumnCount
- widthIncludingColumnHeader

new default value for attribute selectionMode: Multi


file-formats/smbc/type/zif_aff_smbc_v1.intf.abap
```diff
@@ -118,68 +118,84 @@ INTERFACE zif_aff_smbc_v1
...
       "! $minimum 1
       "! $maximum 1000
       "! $default '5'
-      row_count              TYPE i,
+      row_count                     TYPE i,
+      "! <p class="shorttext synchronized">Frozen Column Count</p>
+      "! You can freeze the first columns of a table so that they always remain visible when scrolling the table horizontally
+      "! $minimum 0
+      "! $maximum 30
+      "! $default '0'
```
schneidermic0, line 165: Minor: I guess no default would be needed since 0 is automatically the default value for integers.

file-formats/smbc/type/zif_aff_smbc_v1.intf.abap
```diff
@@ -118,68 +118,84 @@ INTERFACE zif_aff_smbc_v1
...
       "! <p class="shorttext synchronized">Entity Set</p>
-      entity_set             TYPE string,
-      table_type             TYPE ty_table_type,
+      entity_set                    TYPE string,
+      table_type                    TYPE ty_table_type,
       "! <p class="shorttext synchronized">Selection Mode</p>
       "! $values {@link zif_aff_smbc_v1.data:co_selection_mode}
-      "! $default {@link zif_aff_smbc_v1.data:co_selection_mode.auto}
-      selection_mode         TYPE string,
+      "! $default {@link zif_aff_smbc_v1.data:co_selection_mode.multi}
```
schneidermic0, line 145: I think this is an incompatible change! If you really want to change the default, I think you need to introduce a new format version.

Is changing the default really necessary?

schneidermic0 (changes requested):
@sepp4me Thanks for sharing the update. 
Sorry for the late review. I didn't manage to look into it earlier. 

We need to discuss the change of the default value. From my point of view this is an incompatible change which could be covered by a new format version if needed.

Please also fix the ABAP lint issue

file-formats/smbc/type/zif_aff_smbc_v1.intf.abap
sepp4me: @schneidermic0 
The "Auto" setting has been deprecated, see https://sapui5.hana.ondemand.com/sdk/#/topic/116b5d82e8c545e2a56e1b51b8b0a9bd
The new default "Multi" for practical purposes corresponds to the deprecated "Auto". You can compare the description of the "Auto" mode following this link: https://ui5.sap.com/1.96.37/#/topic/116b5d82e8c545e2a56e1b51b8b0a9bd

We are aware that during de-serialization of existing objects the selectionMode can now automatically switch from "Auto" to "Multi"

file-formats/smbc/type/zif_aff_smbc_v1.intf.abap
sepp4me: ABAP lint and default value -> done

file-formats/smbc/type/zif_aff_smbc_v1.intf.abap
schneidermic0: OK, thanks for the update. I understand changing the default from "auto" to "multi" is (semantically) a compatible change.

file-formats/smbc/type/zif_aff_smbc_v1.intf.abap
```diff
@@ -118,68 +118,83 @@ INTERFACE zif_aff_smbc_v1
...
-      hide_column            TYPE ty_personalization-hide_column,
+      hide_column                   TYPE ty_personalization-hide_column,
       "! <p class="shorttext synchronized">Condensed Table Layout</p>
       "! If true, display rows in a condensed way. Only applicaple to Table Type 'GridTable'.
-      condensed_table_layout TYPE abap_bool,
+      condensed_table_layout        TYPE abap_bool,
+      "! <p class="shorttext synchronized">Include column headers in width calculation</p>
+      "! If true, include the column labels while calculating the default column width.
+      "! By default, the column width is calculated based on the type of the content
+      width_including_column_header TYPE abap_bool,
```
schneidermic0, line 188: Minor: Title does not match field name. I don't have a better proposal, but maybe others

schneidermic0 (approved):
In general looks good to me. I am adding UX colleagues.

I see only a mismatch of a field name with its title. But due to length restrictions I also didn't come up with a better proposal

file-formats/smbc/type/zif_aff_smbc_v1.intf.abap
sepp4me: Field name & description are coming from here: https://sapui5.hana.ondemand.com/sdk/#/api/sap.fe.macros.Table%23controlProperties

Field names of the SMBC object are always equal to their FE counterpart attributes to simplify mapping

image

file-formats/smbc/type/zif_aff_smbc_v1.intf.abap
```diff
@@ -118,68 +118,83 @@ INTERFACE zif_aff_smbc_v1
...
+      hide_sort                     TYPE ty_personalization-hide_sort,
       "! <p class="shorttext synchronized">Hide Column</p>
       "! If true, the user cannot add and remove columns to the table
-      hide_column            TYPE ty_personalization-hide_column,
+      hide_column                   TYPE ty_personalization-hide_column,
       "! <p class="shorttext synchronized">Condensed Table Layout</p>
       "! If true, display rows in a condensed way. Only applicaple to Table Type 'GridTable'.
-      condensed_table_layout TYPE abap_bool,
+      condensed_table_layout        TYPE abap_bool,
+      "! <p class="shorttext synchronized">Include column headers in width calculation</p>
```
wurzka, line 185: Title case:
```suggestion
      "! <p class="shorttext synchronized">Include Column Headers in Width Calculation</p>
```

wurzka (changes requested):
Only one problem with the front page, apart from that it looks good to me.

wurzka (approved):
Thanks, looks good to me from UX perspective

file-formats/smbc/type/zif_aff_smbc_v1.intf.abap
sepp4me: done

wurzka:
As changing the default from "auto" to "multi" is (semantically) a compatible change (see [discussion](https://github.com/SAP/abap-file-formats/pull/672#discussion_r1834509016)), we can merge this change without increasing the format version.

PR #673, opened 2024-11-15
[AIFR] - Adding field aifRecipient to AFF for AIFR

D047539:
Short explanation of the background:
During our tests, we discovered the need for an additional field in the AIFR, specifically the AIF Recipient. 
It must be possible to assign the same AIF recipient to different namespaces.

The object name is derived from the namespace and the AIF Recipient. To ensure accurate derivation, we include both fields in the creation wizard and set the object name as read-only.
Since the exact derivation from object name Recipient back to Namespace and AIF Recipient may not always be possible, we displayed both fields as read-only in the editor for informational purposes.

### DIFF
--- a/file-formats/aifr/examples/z_aff_example_aifr.aifr.json
+++ b/file-formats/aifr/examples/z_aff_example_aifr.aifr.json
@@ -6,6 +6,7 @@
     "abapLanguageVersion": "cloudDevelopment"
   },
   "generalInformation":{
-      "namespace": "AIFNS"
+      "namespace": "AIFNS",
+      "aifRecipientName": "REC_WS_INB"
   }
 }
--- a/file-formats/aifr/examples/z_aff_example_aifr.aifr.json
+++ b/file-formats/aifr/examples/z_aff_example_aifr.aifr.json
@@ -7,6 +7,6 @@
   },
   "generalInformation":{
       "namespace": "AIFNS",
-      "aifRecipientName": "REC_WS_INB"
+      "aifRecipient": "REC_WS_INB"
   }
 }
--- a/file-formats/aifr/aifr-v1.json
+++ b/file-formats/aifr/aifr-v1.json
@@ -65,7 +65,7 @@
           "type": "string",
           "maxLength": 15
         },
-        "aifRecipientName": {
+        "aifRecipient": {
           "title": "AIF Recipient",
           "description": "SAP Application Interface recipient",
           "type": "string",
@@ -75,7 +75,7 @@
       "additionalProperties": false,
       "required": [
         "namespace",
-        "aifRecipientName"
+        "aifRecipient"
       ]
     }
   },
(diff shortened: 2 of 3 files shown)


schneidermic0:
Technically, adding new mandatory fields is considered as an incompatible change.

However, since AIFR is still in development, I think we can still accept incompatible changes. This means, we need to ignore the failing compatibility check in this PR.

file-formats/aifr/examples/z_aff_example_aifr.aifr.json
```diff
@@ -6,6 +6,7 @@
     "abapLanguageVersion": "cloudDevelopment"
   },
   "generalInformation":{
-      "namespace": "AIFNS"
+      "namespace": "AIFNS",
+      "aifRecipient": "REC_WS_INB"
```
schneidermic0, line 10: The prefix "AIF" has also been used in other places (like for AIFN for the namespace):
https://github.com/SAP/abap-file-formats/blob/942e346992be50637c8140da6564ddf75f0ef5d2/file-formats/aifn/examples/z_aff_example_aifn.aifn.json#L9

Therefore, I guess it is fine here, too.

schneidermic0 (commented):
In general, the addition looks good to me. I have just added a comment wrt "AIF" prefix. Maybe, others have an opinion on this.

file-formats/aifr/examples/z_aff_example_aifr.aifr.json
D047539: We refer to the development object: Namespace (AIFN). The namespace (AIFN) contains the field AIF namespace, here we basically convert the name of the namespace (AIFN) into our own AIF namespace. This is necessary due to historical reasons.
In this case it is the namespace and not the internal AIF namespace which needs to be assigned to the recipient, so the prefix would not be correct.

schneidermic0: Oh, I see. Sorry, my comment was misleading. 

My comment should have been related to field `aifRecipient` (not to field `namespace`). I just wanted to state, that it is most probably fine to use "aif" as prefix for this field as you did it for `aifNamespace` in AIFN.

wurzka (approved):
We performed UX check on this change. Looks good to me.

file-formats/aifr/examples/z_aff_example_aifr.aifr.json
wurzka: I agree with you. AIF is often used in the editors of this context. Therefore, it is fine for me here as well.

schneidermic0: Fine for me... I resolve this conversation

schneidermic0:
We have skipped the compatibility check. See my comment above

> However, since AIFR is still in development, I think we can still accept incompatible changes. This means, we need to ignore the failing compatibility check in this PR.

PR #675, opened 2024-12-03
[EVTO] Add new object type EVTO

apanys:


file-formats/evto/examples/z_aff_example.evto.json
```diff
@@ -0,0 +1,11 @@
+{
+  "formatVersion": "1",
+  "header": {
+    "description": "Example Event Object.",
+    "originalLanguage": "en"
+  },
+  "majorVersion": 1,
+  "minorVersion": 0,
+  "patchVersion": 4,
+  "sapObjectType": "Z_AFF_EXAMPLE"
```
schneidermic0, line 10: We usually suggest (in the meantime) to not introduce top-level fields. If you don't want to have a specific name for the section you could name it `general_information`.

See https://github.com/SAP/abap-file-formats/blob/main/docs/json.md#writing-json-schema-with-abap-types

schneidermic0, line 10: I am not sure: Does it make sense to move the SAP object type at the first location?

file-formats/evto/examples/z_aff_example.evto.json
```diff
@@ -0,0 +1,11 @@
+{
+  "formatVersion": "1",
+  "header": {
+    "description": "Example Event Object.",
+    "originalLanguage": "en"
+  },
+  "majorVersion": 1,
+  "minorVersion": 0,
+  "patchVersion": 4,
```
schneidermic0, line 9: Just an idea: Another option would be to have something like this, wouldn't it? For me a semantic version is better readable, but you would need to do more validations and parsing it into several ABAP fields...

```suggestion
  "semanticVersion": "1.0.4",
```

What do you think?

schneidermic0 (commented):
Thanks for your contribution. 

I have just checked your example object and added some questions/suggestions

file-formats/evto/examples/z_aff_example.evto.json
apanys: Yes, I will incorporate this suggestion.

apanys: Yes, I will incorporate this suggestion.

file-formats/evto/examples/z_aff_example.evto.json
apanys: Yes, I thought about this but decided against it for consistency. EVTO is quite near to the Event Binding (EVTB) and Service Binding (SRVB) repository objects. Both have the semantic version input fields as separate fields.

file-formats/evto/type/zif_aff_evto_v1.intf.abap
```diff
@@ -0,0 +1,49 @@
...
+    "! General information
+    BEGIN OF ty_general_information,
+      "! <p class="shorttext">SAP Object Type</p>
+      "! SAP Object Type
+      "! $required
+      sap_object_type TYPE c LENGTH 30,
+
+      "! <p class="shorttext">Major Version</p>
+      "! Major version
+      "! $minimum 1
```
schneidermic0, line 15: Just out of curiosity. Version 0.0.1 is not an allowed value?

schneidermic0 (commented):
Thanks for the updates. Looks good to me 👍 

I add the label that the object type is ready for UX review

file-formats/evto/type/zif_aff_evto_v1.intf.abap
apanys: Exactly, we will default it with "1.0.0" and lower than that is not allowed.

apanys:
Thanks a lot :)

wurzka (approved):
Nothing more to add from UX perspective. Looks good to me.

schneidermic0 (approved):

PR #678, opened 2024-12-12
Feature/uiad v2 with dedicated description

HelgeCD:
In the v1 version we did not have a dedicated field for the description in the database. Instead another field was used that is displayed in some scenarios (e.g. smart links, the default application title) in the FLP.

with the v2 version we now have a dedicated field for the description on the database and have therefore added a targetMappingTitle which is the old field which we misused as description.


file-formats/uiad/README.md
schneidermic0, line 1: If a new format version is introduced, we try to keep version v1 in the repository. Could you link to the old version in the readme file.

As reference you could check following readme with 2 format versions:
https://github.com/SAP/abap-file-formats/tree/main/file-formats/cfdf

Maybe, there is one thing I would do differently (compared to CFDF): I would keep the example for the latest version in the existing example (i.e., `z_aff_example_uiad.uiad.json`) and add a `_vx` to the old format version examples (i.e., `z_aff_example_uiad**_v1**.uiad.json`).

schneidermic0 (changes requested):
In general the new version looks good to me. I would suggest to keep the old format version in the repository and adapt the readme file accordingly. 

I think I can also label it as ready for UX review

file-formats/uiad/README.md
HelgeCD: I adjusted it, but now the "Check Markdown Links / Check links (pull_request) " step is failing. From the message it looks like an issue with the step itself and not with our pull request. can you take a look?

schneidermic0: Thanks, @HelgeCD. The broken link issue was caused by an infrastructure issue. See #680.

schneidermic0 (approved):
Looks good to me, now...

wurzka (approved):
From ux-perspective, looks good to me.

PR #690, opened 2025-03-19
Constraints for Defaults in JSON Files

schneidermic0:
Default values defined in AFF interfaces using the annotation `$default` can be problematic if they are optional (non-required) fields and initial values of the underlying ABAP type are allowed values.

Therefore, we need to document some constraints about the default annotation


docs/json.md
```diff
@@ -179,6 +180,13 @@ To set the `default` for a component of the JSON Schema, the annotation
 "! $default 'value'
 ```
 This also ensures that only components whose value is not equal to a specific default value are serialized to the JSON data file.
+
+Default values can only specified if one of the following rules apply:
+
+1. The default value represents the initial value of the underlying data type (e.g., `0` for an interger or `space` for a character field).
+2. The field and all higher-level structures of the field are marked as `$required`.
```
wurzka, line 187: I'm just wondering what about tables? The field with the default must be `$requried` (as well as all structures of higher-level until the table type), but not the table, right? Because here, we have the kind of "null" that means being not an entry.
So I'm not sure if I'm right and if you excluded tables from being `$required` accidentally or intentionally.

wurzka (changes requested):
Some typos and remarks :)

docs/json.md
wurzka: Yes, I agree that it is complex. But I wonder whether that's the reason we should document it (somewhere) so that we do not need to think about it again

schneidermic0: Fixed with https://github.com/SAP/abap-file-formats/pull/690/commits/7c8b44bfa56ac5ecdf1c90d64a7e9fb885025858

wurzka (changes requested):
The part with the special handling for tables looks good to me. See my comments.

docs/json.md
```diff
@@ -179,6 +180,19 @@ To set the `default` for a component of the JSON Schema, the annotation
...
+
+Default values can only specified if one of the following rules apply:
+
+1. The default value represents the initial value of the underlying data type (e.g., `0` for an interger or `space` for a character field).
+
+2. The field and all higher-level structures* of the field are marked as `$required`.
+
+   > *) Arrays don't need to be marked as `$required` in this case.
+   In Json it is possible to distinguish between empty arrays (`"table": []`) and arrays with "empty" array items (`"table": [{}]`).
+   All fields with defaults in the array items that don't follow the other rules, along with their higher-level structures, must be marked as `$required`.
```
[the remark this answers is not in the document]
schneidermic0: If rule 3 is followed (i.e., an enum has a default value that is not the initial value of the underlying type, but there is no other (non-default) enum value that represents this initial value), we won't have any issue, right?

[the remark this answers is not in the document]
schneidermic0: Nevertheless, would it make sense to exchange rule 2 and 3?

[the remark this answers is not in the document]
wurzka: Yes, you're right. In this case, we can distinguish between default and not set. But I agree with changing the order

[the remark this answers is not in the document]
schneidermic0: Changed order with https://github.com/SAP/abap-file-formats/pull/690/commits/af524ceb06ec37c21eecf09f2d98cfb4fc275f8b

docs/json.md
```diff
@@ -179,6 +180,19 @@ To set the `default` for a component of the JSON Schema, the annotation
 "! $default 'value'
 ```
 This also ensures that only components whose value is not equal to a specific default value are serialized to the JSON data file.
+
+Default values can only specified if one of the following rules apply:
```
wurzka, line 184: I guess I missed this in my last review, but `be` has vanished
```suggestion
Default values can only be specified if one of the following rules apply:
```

wurzka (changes requested):
Just one minor, else looks good to me

docs/json.md
schneidermic0: This was a regression, again :/

wurzka (approved):
Thanks, looks good to me

PR #700, opened 2025-04-04
Add Object Type Descriptions to `README` Files

### LINKED ISSUE #698: Add Description for Object Types in the Description
It would be nice to have also a readable description for the object type in the README files for each object type.

Something like:

- INTF - Interfaces
- CLAS - Classes
- AIFP - Application Interface Checks
- ...

schneidermic0:
Closes #698


albertmink:
Do we need the section `File Structure` if some few objects make us of?
Other than that, I like the change

schneidermic0:
> Do we need the section `File Structure` if some few objects make us of?

Do you mean that most object types only have one JSON file? I would still like to keep the table with the links to the schema and example .... That makes it more convenient to navigate. Or do you struggle with the header?

> Do we need to adopt the generation report over [here](https://github.com/SAP/abap-file-formats-tools/blob/2e342ea9dab6c5c487093342e509e411cb1eddbf/src/z_generate_repo.prog.abap#L391C21-L398C21) to prevent future manual activities?

Yes, we should adapt the program, too. And all open PRs that touch the README

albertmink:
> > Do we need the section `File Structure` if some few objects make us of?
> 
> Do you mean that most object types only have one JSON file? I would still like to keep the table with the links to the schema and example .... That makes it more convenient to navigate. Or do you struggle with the header?
>

Omg, I didn't check the entire file. In the PR is looked to me as if the table is empty. But it is not 🙈

albertmink (approved):
LGTM

wurzka (changes requested):
Found a typo ;)

schneidermic0:
- [ ] I need to update this PR to add the descriptions for AIFP that has just been merged

GuilhermeSaraiva96 (approved):

Issue #697, opened 2025-04-02
RVBC: Default value for enum `ty_extensibility_mode` is not the initial value


schneidermic0:
For object type RVBC, the default value for enum `ty_extensibility_mode` is not the initial value.

Refer to https://github.com/SAP/abap-file-formats/blob/4df86464e64f3e7e3fe14ee23c23075d8abf4366/file-formats/rvbc/type/zif_aff_rvbc_v1.intf.abap#L9

Documentation see https://github.com/SAP/abap-file-formats/blob/main/docs/json.md#default-values.

schneidermic0:
I see two options to solve the issue:

1. Changing the default value to `fully supported`
2. Mark the fields `booklet_definition` and its sub field `extensibility_mode` as `$required`

I think both options are not the best, but I don't see a better proposal :(

Furthermore, both changes are potentially incompatible.

schneidermic0:
I have just informed the object type owners of RVBC

RuixingYangSAP:
Hi @schneidermic0, I have made the changes and created the pull request: https://github.com/SAP/abap-file-formats/pull/701. however there were some check failed which I could not really understand the root cause. Could you please kindly help have a look? Thx & best regards, Ray

schneidermic0:
This was solved by #701

PR #701, opened 2025-04-09
[RVBC] Add $default to extensibility_mode and booklet_definition

RuixingYangSAP:


schneidermic0:
Fixes #697

file-formats/rvbc/type/zif_aff_rvbc_v1.intf.abap
```diff
@@ -62,6 +62,7 @@ INTERFACE zif_aff_rvbc_v1
       application        TYPE ty_app_id,
       "! <p class="shorttext">Extensibility Mode</p>
       "! Extensibility mode
+      "! $required
```
schneidermic0, line 65: Sorry, my suggestion [here](https://github.com/SAP/abap-file-formats/issues/697#issuecomment-2773070415) was just wrong/misleading

I think extensibility mode doesn't need to be `$required`. It would be sufficient to mark `booklet_definition` as required.

See rule 3 in [documentation](https://github.com/SAP/abap-file-formats/blob/main/docs/json.md#default-values):

> 3. All higher-level structures* of the field are marked as `$required`. 

So, if you like, we could remove the the annotation here, again.

schneidermic0 (commented):
@RuixingYangSAP Thanks for taking this up.

Sorry, I think I made a wrong suggestion in the issue. We could solve the issue even in a simpler way with marking just `booklet_definition` as required. I am sorry about any confusion that my suggestion might have caused.

file-formats/rvbc/type/zif_aff_rvbc_v1.intf.abap
RuixingYangSAP: HI @schneidermic0 , sure, I just removed the $required on the field extensibility_mode. somehow the new build still has failed Compability Check. Is that something I can fix from my side? BR, Ray

schneidermic0 (approved):
Thanks, @RuixingYangSAP for fixing this. It looks good to me.

> somehow the new build still has failed Compability Check. Is that something I can fix from my side?

Changing a field to required is technically an incompatible change. This is what the check reports. As mentioned before, I believe we can ignore it, because RVBC hasn't been used yet.

RuixingYangSAP:
> Thanks, @RuixingYangSAP for fixing this. It looks good to me.
> 
> > somehow the new build still has failed Compability Check. Is that something I can fix from my side?
> 
> Changing a field to required is technically an incompatible change. This is what the check reports. As mentioned before, I believe we can ignore it, because RVBC hasn't been used yet.

@schneidermic0  Thanks for explaining it. The last quesion: Would you take care merging the pull request please?

Christopher-Hermann (approved):

schneidermic0:
We must ignore the compatibility check in this case.

schneidermic0:
@RuixingYangSAP Thanks for your help fixing this issue.

PR #703, opened 2025-04-15
[AIFA] Add new object type

Smartynkov:


file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
```diff
@@ -0,0 +1,102 @@
...
+      "! Namespace
+      "! $required
+      namespace      TYPE c LENGTH 15,
+      "! <p class="shorttext">AIF Action</p>
+      "! AIF action
+      "! $required
+      aif_action     TYPE c LENGTH 20,
+      "! <p class="shorttext">Main Component Type</p>
+      "! Main component type
+      main_comp_type TYPE zif_aff_types_v1=>ty_object_name_30,
```
GuilhermeSaraiva96, line 44:
```suggestion
      main_component_type TYPE zif_aff_types_v1=>ty_object_name_30,
```

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
```diff
@@ -0,0 +1,102 @@
...
+    END OF co_error_handling.
+
+  TYPES:
+    "! $values { @link zif_aff_aifa_v1.data:co_error_handling }
+    "! $default { @link zif_aff_aifa_v1.data:co_error_handling.ignore_data }
+    error_handling_type TYPE c LENGTH 1,
+
+    "! <p class="shorttext">Field To Check</p>
+    "! Field to check
+    BEGIN OF ty_fields_check,
```
GuilhermeSaraiva96, line 21:
```suggestion
    BEGIN OF ty_field_to_check,
```

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
```diff
@@ -0,0 +1,102 @@
...
+      "! <p class="shorttext">Field Name</p>
+      "! Field name
+      field_name TYPE c LENGTH 255,
+    END OF ty_fields_check,
+    "! <p class="shorttext">Fields To Restore</p>
+    "! Fields to restore
+    tty_fields_check TYPE STANDARD TABLE OF ty_fields_check WITH DEFAULT KEY.
```
GuilhermeSaraiva96:
```suggestion
    ty_fields_to_check TYPE STANDARD TABLE OF ty_fields_check WITH DEFAULT KEY.
```

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
```diff
@@ -0,0 +1,102 @@
...
+      "! $required
+      check_obj_name TYPE zif_aff_types_v1=>ty_object_name_30,
+      "! <p class="shorttext">Check Behavior</p>
+      "! Check behavior
+      check_ignore   TYPE error_handling_type,
+      "! <p class="shorttext">Field Name</p>
+      "! Field name
+      field_name     TYPE tty_fields_check,
+    END OF ty_check_data,
+    tty_check_data TYPE STANDARD TABLE OF ty_check_data WITH DEFAULT KEY,
```
GuilhermeSaraiva96:
```suggestion
    ty_checks TYPE STANDARD TABLE OF ty_check_data WITH DEFAULT KEY,
```

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
```diff
@@ -0,0 +1,102 @@
...
+      "! Main component type
+      main_comp_type TYPE zif_aff_types_v1=>ty_object_name_30,
+      "! <p class="shorttext">Class</p>
+      "! Class
+      class          TYPE zif_aff_types_v1=>ty_object_name_30,
+    END OF ty_general_information,
+
+    "! <p class="shorttext">Check Data</p>
+    "! Check data
+    BEGIN OF ty_check_data,
```
GuilhermeSaraiva96:
```suggestion
    BEGIN OF ty_check,
```

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
```diff
@@ -0,0 +1,102 @@
...
+      "! <p class="shorttext">Number</p>
+      "! Number
+      "! $required
+      number         TYPE n LENGTH 3,
```
GuilhermeSaraiva96: What does number stand for? Is it an index? Or the order of the checks to be performed?

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
```diff
@@ -0,0 +1,102 @@
...
+      "! Header
+      "! $required
+      header              TYPE zif_aff_types_v1=>ty_header_60_cloud,
+      "! <p class="shorttext">General Information</p>
+      "! General information
+      "! $required
+      general_information TYPE ty_general_information,
+      "! <p class="shorttext">Fields To Restore Assignment</p>
+      "! Fields to restore assignment
+      field_restore       TYPE tty_fields_restore,
```
GuilhermeSaraiva96:
```suggestion
      fields_to_restore       TYPE tty_fields_restore,
```

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
```diff
@@ -0,0 +1,102 @@
...
+      format_version      TYPE zif_aff_types_v1=>ty_format_version,
+      "! <p class="shorttext">Header</p>
+      "! Header
+      "! $required
+      header              TYPE zif_aff_types_v1=>ty_header_60_cloud,
+      "! <p class="shorttext">General Information</p>
+      "! General information
+      "! $required
+      general_information TYPE ty_general_information,
+      "! <p class="shorttext">Fields To Restore Assignment</p>
```
GuilhermeSaraiva96, line 96:
```suggestion
      "! <p class="shorttext">Fields To Restore</p>
```

Assignment in this context does not provide extra information in my opinion

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
```diff
@@ -0,0 +1,102 @@
...
+      "! Fields to restore assignment
```
GuilhermeSaraiva96, line 97:
```suggestion
      "! Fields to restore
```

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
```diff
@@ -0,0 +1,102 @@
...
+      field_restore       TYPE tty_fields_restore,
+      "! <p class="shorttext">Checks Assignment</p>
```
GuilhermeSaraiva96, line 99:
```suggestion
      "! <p class="shorttext">Checks</p>
```

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
```diff
@@ -0,0 +1,102 @@
...
+      "! Checks assignment
+      check_data          TYPE tty_check_data,
```
GuilhermeSaraiva96:
```suggestion
      checks          TYPE tty_check_data,
```

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
```diff
@@ -0,0 +1,102 @@
...
+      header              TYPE zif_aff_types_v1=>ty_header_60_cloud,
+      "! <p class="shorttext">General Information</p>
+      "! General information
+      "! $required
+      general_information TYPE ty_general_information,
+      "! <p class="shorttext">Fields To Restore Assignment</p>
+      "! Fields to restore assignment
+      field_restore       TYPE tty_fields_restore,
+      "! <p class="shorttext">Checks Assignment</p>
+      "! Checks assignment
```
GuilhermeSaraiva96, line 100:
```suggestion
      "! Checks
```

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
```diff
@@ -0,0 +1,102 @@
...
+      check_ignore   TYPE error_handling_type,
+      "! <p class="shorttext">Field Name</p>
+      "! Field name
+      field_name     TYPE tty_fields_check,
+    END OF ty_check_data,
+    tty_check_data TYPE STANDARD TABLE OF ty_check_data WITH DEFAULT KEY,
+
+    "! <p class="shorttext">Field To Restore</p>
+    "! Field to restore
+    BEGIN OF ty_fields_restore,
```
GuilhermeSaraiva96:
```suggestion
    BEGIN OF ty_field_to_restore,
```

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
```diff
@@ -0,0 +1,102 @@
...
+      "! <p class="shorttext">Field Name</p>
+      "! Field name
+      field_name TYPE c LENGTH 255,
```
GuilhermeSaraiva96, line 77:
```suggestion
      name TYPE c LENGTH 255,
```

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
```diff
@@ -0,0 +1,102 @@
...
+
+  TYPES:
+    "! $values { @link zif_aff_aifa_v1.data:co_error_handling }
+    "! $default { @link zif_aff_aifa_v1.data:co_error_handling.ignore_data }
+    error_handling_type TYPE c LENGTH 1,
+
+    "! <p class="shorttext">Field To Check</p>
+    "! Field to check
+    BEGIN OF ty_fields_check,
+      "! <p class="shorttext">Field Name</p>
```
GuilhermeSaraiva96, line 22: WHat is the relationship between this Field and the Check? Do Checks contain fields or the other way around?

Smartynkov: Hi, in Action we have a numbered list of fields that are passed to Check. Check contains a condition which fields should be checked something like: 'ID = $1 name = $2' where $1 and $2 are the first and second fields in the Action.

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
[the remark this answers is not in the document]
Smartynkov: Its string now.

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
Smartynkov: This number shows the order in which checks will be performed.

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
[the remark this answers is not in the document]
Smartynkov: IN the future it will be field path. In case we have a deep structure with many substructures the path will be longer than char 40.

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
GuilhermeSaraiva96: missing

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
```diff
@@ -0,0 +1,103 @@
...
+    "! General information
+    BEGIN OF ty_general_information,
+      "! <p class="shorttext">Namespace</p>
+      "! Namespace
+      "! $required
+      namespace           TYPE c LENGTH 15,
+      "! <p class="shorttext">AIF Action</p>
+      "! AIF action
+      "! $required
+      aif_action          TYPE c LENGTH 20,
```
GuilhermeSaraiva96, line 40: Is the aif_action the name of the object instance itself?

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
```diff
@@ -0,0 +1,103 @@
...
+    "! Check details
+    BEGIN OF ty_check_details,
+      "! <p class="shorttext">Number</p>
+      "! Number
+      "! $required
+      number          TYPE n LENGTH 3,
+      "! <p class="shorttext">Check</p>
+      "! Check
+      "! $required
+      check_obj_name  TYPE zif_aff_types_v1=>ty_object_name_30,
```
GuilhermeSaraiva96, line 59:
```suggestion
      check  TYPE zif_aff_types_v1=>ty_object_name_30,
```

It is implicit that refers to the name of the check object

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
```diff
@@ -0,0 +1,103 @@
...
+      "! <p class="shorttext">Check Behavior</p>
+      "! Check behavior
+      check_behaviour TYPE ty_error_handling_type,
+      "! <p class="shorttext">Field Name</p>
+      "! Field name
```
GuilhermeSaraiva96, line 64:
```suggestion
      "! <p class="shorttext">Fields</p>
      "! Fields
```

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
```diff
@@ -0,0 +1,103 @@
...
+      field_name      TYPE ty_fields_to_check,
+    END OF ty_check_details,
+    "! <p class="shorttext">Check Data</p>
+    "! Check data
```
GuilhermeSaraiva96, line 68:
```suggestion
    "! <p class="shorttext">Checks</p>
    "! Checks
```

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
```diff
@@ -0,0 +1,103 @@
...
+      "! <p class="shorttext">General Information</p>
+      "! General information
+      "! $required
+      general_information TYPE ty_general_information,
+      "! <p class="shorttext">Fields To Restore</p>
+      "! Fields to restore
+      field_restore       TYPE ty_fields_to_restore,
+      "! <p class="shorttext">Checks</p>
+      "! Checks
+      check_data          TYPE ty_checks,
```
GuilhermeSaraiva96, line 100:
```suggestion
      checks          TYPE ty_checks,
```

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
```diff
@@ -0,0 +1,103 @@
...
+      "! Header
+      "! $required
+      header              TYPE zif_aff_types_v1=>ty_header_60_cloud,
+      "! <p class="shorttext">General Information</p>
+      "! General information
+      "! $required
+      general_information TYPE ty_general_information,
+      "! <p class="shorttext">Fields To Restore</p>
+      "! Fields to restore
+      field_restore       TYPE ty_fields_to_restore,
```
GuilhermeSaraiva96, line 97:
```suggestion
      fields_to_restore       TYPE ty_fields_to_restore,
```

GuilhermeSaraiva96, line 97: What are these fields to restore? How do they fit into the workflow of the AIF Action?

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
```diff
@@ -0,0 +1,103 @@
...
+      class               TYPE zif_aff_types_v1=>ty_object_name_30,
+    END OF ty_general_information,
+
+    "! <p class="shorttext">Check Details</p>
+    "! Check details
+    BEGIN OF ty_check_details,
+      "! <p class="shorttext">Number</p>
+      "! Number
+      "! $required
+      number          TYPE n LENGTH 3,
```
GuilhermeSaraiva96, line 55: The table is of type standard, which means that the order of the items in the table define a sequence, which can be used to derive this 'number'. Basically, the first item in the list would be the first check to be performed. In this sense, the 'number' field here be useless. This also makes it more intuitive to use in our Editors. What do you think?

GuilhermeSaraiva96 (changes requested):

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
Smartynkov: Hi, I discussed with colleagues, and we agreed that we can remove the number field from the UI.

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
Smartynkov: Sometimes, later action classes depend on information calculated by earlier action classes. If a later action class fails and the system restarts, an earlier action class might not run again, causing the needed information to be lost. To prevent errors in the restarted process, a "field to restore" can be used. This field saves the important information, so if an action class needs to be re-run after an error, it can retrieve the saved information and continue correctly.

Smartynkov:
Hi, we had a long discussion with a team about changes with Action file format and here is the summary:

We need number id because it's a key field for our table and its shouldn't be calculated automatically. Earlier it was a key and an order number of a check, we decided to spilt t up and now number field isn't about an order of a check and just a key field. we have a new field for the order, and it will be defined by the order of a check in the table on the UI we will add order changeable to the table for that to happen. 

Also, there was some discussion about 'number' as a name for this key field on the UI and we think it's not showing what is this field is about. We propose name 'numeric_id', but we are also opened for the discussion.

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
GuilhermeSaraiva96: Can't another field be the key? For example, the check name. Are you expecting multiple entries with the same check name?

From my point of view, I cannot see the advantage in you keeping this numeric_id, as with the table type standard, there is already an implied id in each data entry.

Smartynkov: Hi, actually we do expect multiple entries with the same check name. We can do same check for few fields. And Field may be longer than 128, because its field path actually. So, 'field path' can't be the key as well. We need number as a key.

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
Smartynkov: Its Action's name, but Action object name forms from several fields, including aif_action.

file-formats/aifa/README.md
schneidermic0, line 1: Example Object is missing. I guess it would be really helpful even for the review to have an example object to see some potential data.

If you cannot provide an example, yet, this is just a reminder, to create an issue in this repo, before we merge the change.

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
```diff
@@ -0,0 +1,103 @@
...
+      "! <p class="shorttext">AIF Action</p>
+      "! AIF action
+      "! $required
+      aif_action          TYPE c LENGTH 20,
+      "! <p class="shorttext">Main Component Type</p>
+      "! Main component type
+      main_component_type TYPE zif_aff_types_v1=>ty_object_name_30,
+      "! <p class="shorttext">Class</p>
+      "! Class
+      class               TYPE zif_aff_types_v1=>ty_object_name_30,
```
schneidermic0, line 46: I guess we are talking about an ABAP class. What is the purpose of the class? Would it make sense to add the purpose to the field name?

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
```diff
@@ -0,0 +1,103 @@
...
+   PUBLIC.
+
+  "! $values { @link zif_aff_aifa_v1.data:co_error_handling }
+  "! $default { @link zif_aff_aifa_v1.data:co_error_handling.ignore_data }
+  TYPES ty_error_handling_type TYPE c LENGTH 1.
+  CONSTANTS:
+    BEGIN OF co_error_handling,
+      "! <p class="shorttext">Treat As Error</p>
+      "! Treat as error if check is not successful
+      trigger_error TYPE c LENGTH 1 VALUE '',
```
schneidermic0, line 11: Text and file name do not match. I guess I would try avoid to use "trigger" if possible since it shall be avoided according to https://help.sap.com/docs/TERMINOLOGY/25cbeaaad3c24eba8ea10b579ce81aa1/83a23df24013403ea4c1fdd0107cc0fd.html

However, "Treat as Error" or "Raise Error" should be fine.

```suggestion
      "! <p class="shorttext">Treat as Error</p>
      "! Treat as error if check is not successful
      treat_as_error TYPE c LENGTH 1 VALUE '',
```

```suggestion
      "! <p class="shorttext">Raise Error</p>
      "! Raise error if check is not successful
      raise_error TYPE c LENGTH 1 VALUE '',
```

schneidermic0 (changes requested):
Thanks for your proposals, @Smartynkov. @GuilhermeSaraiva96 asked me to review the changes, too. I have added some questions and comments.

I guess an example object would also help to understand your use case even better.

file-formats/aifa/README.md
```diff
@@ -0,0 +1,11 @@
+# Action File Format
```
albertmink, line 1: Observations related to recent README updates: @schneidermic0 this section title is generated from the current SVAL object. Merging the pr then may lead to uncorrect README as the SVAL will be checked much later in our process.

GuilhermeSaraiva96:
> Hi, we had a long discussion with a team about changes with Action file format and here is the summary:
> 
> We need number id because it's a key field for our table and its shouldn't be calculated automatically. Earlier it was a key and an order number of a check, we decided to spilt t up and now number field isn't about an order of a check and just a key field. we have a new field for the order, and it will be defined by the order of a check in the table on the UI we will add order changeable to the table for that to happen.
> 
> Also, there was some discussion about 'number' as a name for this key field on the UI and we think it's not showing what is this field is about. We propose name 'numeric_id', but we are also opened for the discussion.

Hi,

We have discussed this in our team. We have a few questions:
1.You mentioned that the key field is not calculated automatically. How will the user know what to write then? Does it have to be a random number so that it does not collide with the other ids?
2. Is the check field also a key field, together with numeric id?
3. What do you think of have that id as a textual id instead of a number? It would be kind of a normal id, so that it also has some meaning and makes sense to the user

GuilhermeSaraiva96:
> > > Hi, we had a long discussion with a team about changes with Action file format and here is the summary:
> > > We need number id because it's a key field for our table and its shouldn't be calculated automatically. Earlier it was a key and an order number of a check, we decided to spilt t up and now number field isn't about an order of a check and just a key field. we have a new field for the order, and it will be defined by the order of a check in the table on the UI we will add order changeable to the table for that to happen.
> > > Also, there was some discussion about 'number' as a name for this key field on the UI and we think it's not showing what is this field is about. We propose name 'numeric_id', but we are also opened for the discussion.
> > 
> > 
> > Hi,
> > We have discussed this in our team. We have a few questions: 1.You mentioned that the key field is not calculated automatically. How will the user know what to write then? Does it have to be a random number so that it does not collide with the other ids? 2. Is the check field also a key field, together with numeric id? 3. What do you think of have that id as a textual id instead of a number? It would be kind of a normal id, so that it also has some meaning and makes sense to the user
> 
> Hi, I will attempt to answer the questions, but I recommend scheduling a brief meeting to discuss them further. Regarding question 1: In the past, the customer would enter a sequence number that also functioned as an identification number. We now aim to split this dual-purpose by separating the sequence number from the identification number. Consequently, the customer will now be required to input a unique 3-digit identification number. To question 2: No, the check field is not part of the key because it can be assigned multiple times with different validation parameters. To question 3: Because this is legacy functionality, and the number is already included in the key, we are unable to make changes to it. Moreover, some existing tables have rather lengthy keys, which unfortunately limits the available space for using text fields as keys.

I will schedule a brief meeting with us and Michael then.
To question 2: the check could still be part of the key. Basically, the check would be key 1 and to separate the entries with different validation parameters, you could use this numeric_id as a second key.
To question 3: I hope I understood the problem well. Could it be that the new values could still be textual, and the old ones stay as numeric? The datatype would basically be string, so both types are supported.
Since you are the experts of the domain and we are not, sometimes we oversee details or present solutions that are not feasible. We are insisting on this topic because it we think it could be improved, maybe it just cannot :)

file-formats/aifa/README.md
schneidermic0: Yeah, on the other hand, I think it is helpful to see a meaningful name (even if it is not 100% correct and might be updated in the future) instead of having none.

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
schneidermic0: I prefer some addition like it is done here:
https://github.com/SAP/abap-file-formats/blob/158e208851c97f795e85738b43fc1e761b18c626/file-formats/chko/examples/z_aff_example_chko.chko.json#L8

However, I have just checked the repository. For AIFP, also "class" was used. So this is kind of consistent at least in your domain:
https://github.com/SAP/abap-file-formats/blob/158e208851c97f795e85738b43fc1e761b18c626/file-formats/aifp/aifp-v1.json#L377

schneidermic0 (approved):
It seems to me the suggestions/comments I have made have been improved with the latest commits.

It looks good to me. Thanks, @Smartynkov

file-formats/aifa/type/zif_aff_aifa_v1.intf.abap
wurzka: I reopened this discussion since the issue also came up during UX review. We suggest to use 
```suggestion
      "! <p class="shorttext">Implementing Class</p>
      "! Implementing Class
      implementing_class               TYPE zif_aff_types_v1=>ty_object_name_30,
```
In my opinion, this is different to AIFP here, since there, `class` is part of the component `Custom Implementation`. Here, it is in `General Information`. So in my opinion, the purpose of the class would be helpful.

wurzka (changes requested):
Please see my comment.

PR #704, opened 2025-04-24
[RVBC] add $required to source and status fields

RuixingYangSAP:
add $required to source and status fields to ensure them appear in the JSON serialization


RuixingYangSAP:
HI @schneidermic0 , as we discussed yesterday, I am creating the PR to synchronize the AFF. Could you please kindly have a look and approve? Thx!

schneidermic0:
@RuixingYangSAP marked fields `source` and `status` as required so that also default values of the enum are serialized furthermore it seems to make sense to specify these values anyhow.

Even in the example file these default values were rendered as of today: 

https://github.com/SAP/abap-file-formats/blob/3a1ade1afe80a23baa4794c1481b7c1bc3b50f2a/file-formats/rvbc/examples/z_aff_example_rvbc.rvbc.json#L8-L9

schneidermic0 (approved):
Thanks for the update, @RuixingYangSAP. Looks good to me. The example is already up-to-date with the existing change.

We can skip the compatibility error here. The finding is correct. It's an incompatible change. However, the AFF has not been used, yet, because it is still in development.

@SAP/abap-file-formats-team Would someone else review it, too?

Markus1812 (approved):
👍

PR #717, opened 2025-08-28
[AIFA] Changes in AFF

Smartynkov:
Hi, the purpose of this change is to improve the user experience with value helps by separating the combined name/value field into two distinct columns: one for the Field Name and another for the Value. The reason of this change is that we cant provide the value help for the constant values and user need to input them manually, so we separate it.

### DIFF
--- a/generate/package-lock.json
+++ b/generate/package-lock.json
@@ -9,38 +9,38 @@
       "version": "0.0.0",
       "license": "MIT",
       "dependencies": {
-        "@abaplint/database-sqlite": "^2.10.24",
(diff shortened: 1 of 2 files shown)


wurzka:
Now, only the compatibility check is failing. This is because a new field is added which might be incompatible. Since AIFA still is in development, we can ignore this check.

Smartynkov:
> Now, only the compatibility check is failing. This is because a new field is added which might be incompatible. Since AIFA still is in development, we can ignore this check.

Thanks a lot! now I am waiting for approval I assume?

PR #719, opened 2025-09-17
DDLS: Adding default value for sourceOrigin

alexanderKononov-sap:


Markus1812 (approved):
Thanks for the update

PR #720, opened 2025-09-30
File Format for MSAG (Message Class)

yanapu-jyothi:
Long text support is not currently included in the ABAP file format. However, support for long text will be added in a future version.


file-formats/msag/type/zif_aff_msag_v1.intf.abap
```diff
@@ -0,0 +1,36 @@
...
+  PUBLIC.
+
+  TYPES:
+    "! <p class="shorttext">Message Properties</p>
+    "! Message properties
+    BEGIN OF ty_message,
+      "! <p class="shorttext">Message Number</p>
+      "! Message number
+      number TYPE c LENGTH 3,
+      "! <p class="shorttext">Short Text</p>
```
schneidermic0: Really minor. I would use only "Text" or "Message Text" for the tile.

Issue #130, opened 2021-07-20
File Formats for Documentation


schneidermic0:
We have to define how we represent documentation (especially SAPscript) in ABAP file formats.

Questions are from my point of view:

1. Technical (using two columns like in classic SAPscript editor) or representation in a dedicated file format which is close to documentation
2. Is the documentation part of the Json or dedicated files (relates also to #115)

schneidermic0:
Actually, I would see documentation in dedicated files. Documentation is translation relevant (#106). Therefore, I think we need to include file name additions for the language as described [here](https://github.com/SAP/abap-file-formats/blob/main/doc/specification.md#language).

PR #743, opened 2026-03-13
[DOMA] add longtexts

larshp:
example in https://github.com/abapGit-tests/DOMA-longtext/blob/main/src/zdoma_long.doma.xml

also relevant for MSAG, https://github.com/SAP/abap-file-formats/blob/main/file-formats/msag/README.md

found out while working on https://github.com/abapGit/abapGit/pull/7606


larshp:
I think lets move forward

#130 was opened 5 years ago

schneidermic0:
I am sorry for the late response. I have already got feedback by the colleagues some days ago but I didn't manage to answer here, earlier.

There is a team working on another, more readable representation for SAP Script documentation, currently. I cannot share more details on this for now.

I see two options for going forward for the integration in abapGit:

1. Use the proposal in this PR, for going forward and adapt it to the new representation as soon as it is available. This requires further adaptations (and maybe, compatibility handling).
2. Don't change the format for documentation for domains (and further objects) and stick to the existing abapGit representation. Then adapt to the new format as soon as it is available.

In any case, I would suggest to move the documentation to a dedicated file instead of incorporating it into the existing JSON representation. 

What do you think?

larshp:
-> single file as json will work for now -> PR needs a bit of rework and defining some structure

file-formats/zif_aff_longtexts_types_v1.intf.abap
```diff
@@ -0,0 +1,33 @@
+"! <p class="shorttext synchronized" lang="en">Long text types</p>
+"! Types for AFF long texts
+INTERFACE zif_aff_longtexts_types_v1 PUBLIC.
```
schneidermic0: LIMU **DOCU** could end up in:

```suggestion
INTERFACE zif_aff_docu_v1 PUBLIC.
```

file-formats/doma/README.md
```diff
@@ -11,3 +11,4 @@ DOMA  | Domain | Dictionary
 File | Cardinality | Definition | Schema | Example
 :--- | :--- | :--- | :--- | :---
 `<name>.doma.json` | 1 | [`zif_aff_doma_v1.intf.abap`](./type/zif_aff_doma_v1.intf.abap) | [`doma-v1.json`](./doma-v1.json) | [`z_aff_example.doma.json`](./examples/z_aff_example.doma.json)
+`<name>.doma.longtext.json` | 0...1 | [`zif_aff_longtexts_types_v1.intf.abap`](../zif_aff_longtexts_types_v1.intf.abap) | [`longtexts-v1.json`](../longtexts-v1.json) | [`z_aff_example.doma.longtext.json`](./examples/z_aff_example.doma.longtext.json)
```
schneidermic0: This is a minor: And for sure we can debate about this: I suggest to use either "documentation" or just "docu" instead of "longtext". This fits also better to the transport object LIMU DOCU. Longtext might to fit for domains for others I believe docu(mentation) would be better.

What do you think?

schneidermic0 (commented):
Thanks for splitting this up into a dedicated file. This make sense to me.

Format in general also looks good to me. I have just some naming proposals to avoid "longtexts".

file-formats/zif_aff_longtexts_types_v1.intf.abap
larshp: updated

file-formats/doma/README.md
larshp: updated

larshp:
@schneidermic0 comments above addressed

file-formats/zif_aff_docu_v1.intf.abap
```diff
@@ -0,0 +1,33 @@
...
+      "! $required
+      line   TYPE c LENGTH 72,
+    END OF ty_line,
+    "! <p class="shorttext">Long Text Lines</p>
+    "! Long text lines
+    ty_lines TYPE STANDARD TABLE OF ty_line WITH DEFAULT KEY.
+
+  TYPES:
+    "! <p class="shorttext">ABAP Daemon</p>
+    "! ABAP daemon
```
schneidermic0:
```suggestion
    "! <p class="shorttext">Documentation</p>
    "! Documentation
```

file-formats/docu-v1.json
```diff
@@ -0,0 +1,49 @@
+{
+  "$comment": "This file is autogenerated, do not edit manually, see https://github.com/SAP/abap-file-formats for more information.",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://github.com/SAP/abap-file-formats/blob/main/file-formats/longtexts-v1.json",
+  "title": "ABAP Daemon",
+  "description": "ABAP daemon",
```
schneidermic0:
```suggestion
  "title": "Documentation",
  "description": "documentation",
```

schneidermic0 (commented):
Sorry, I missed the update. Thanks for pinging me.

I just find a very small thing that I missed before. I try to include the suggested changes right away.

schneidermic0 (approved):

file-formats/doma/README.md
```diff
@@ -11,3 +11,4 @@ DOMA  | Domain | Dictionary
 File | Cardinality | Definition | Schema | Example
 :--- | :--- | :--- | :--- | :---
 `<name>.doma.json` | 1 | [`zif_aff_doma_v1.intf.abap`](./type/zif_aff_doma_v1.intf.abap) | [`doma-v1.json`](./doma-v1.json) | [`z_aff_example.doma.json`](./examples/z_aff_example.doma.json)
+`<name>.doma.docu.json` | 0...1 | [`zif_aff_docu_v1.intf.abap`](../zif_aff_docu_v1.intf.abap) | [`docu-v1.json`](../docu-v1.json) | [`z_aff_example.doma.docu.json`](./examples/z_aff_example.doma.docu.json)
```
albertmink, line 14: file name is good enough till there is an object type `docu` ;)

file-formats/docu-v1.json
```diff
@@ -0,0 +1,50 @@
...
+  "title": "ABAP Daemon",
+  "title": "Documentation",
+  "description": "documentation",
+  "type": "object",
+  "properties": {
+    "formatVersion": {
+      "title": "ABAP File Format Version",
+      "description": "The ABAP file format version",
+      "type": "string",
+      "const": "1"
```
albertmink, line 14: There will be a time, where neither doma is V1 nor the documentation. We then need to ask ourselves what does the versioning mean (same for translations). Didn't blocked us to do the translation, hence shouldn't block here either.

file-formats/doma/README.md
schneidermic0: There will be a (LIMU) object type docu. But then we could use this interface, can't we?

schneidermic0: If we don't think this is an issue, I think we can merge this.

larshp:
@schneidermic0 new changes require another approval, looks like merges does that too

albertmink (approved):
LGTM

schneidermic0:
> new changes require another approval, looks like merges does that too

I didn't find a setting for the automatic merges. Do you know where this can be specified?

PR #745, opened 2026-03-17
[ABIQ] Add new object type ABIQ

LutzR-SAP:


Markus1812 (commented):
Thanks for contributing this new object type! I went through the interface definition and have a few suggestions to align it with AFF conventions. Most are small fixes — the main ones are a copy-paste error in `co_output_type` and `$required` on `abap_bool` fields which is not allowed.

file-formats/abiq/type/zif_aff_abiq_v1.intf.abap
```diff
@@ -0,0 +1,149 @@
...
+  "! $values { @link zif_aff_abiq_v1.data:co_quick_action_type }
+  "! $default { @link zif_aff_abiq_v1.data:co_quick_action_type.llm }
+  TYPES ty_quick_action_type TYPE n LENGTH 2.
+  CONSTANTS:
+    BEGIN OF co_quick_action_type,
+      "! <p class="shorttext">LLM</p>
+      "! LLM Quick Action
+      llm  TYPE ty_quick_action_type VALUE 1,
+      "! <p class="shorttext">JSON</p>
+      "! JSON Quick Action
```
Markus1812:
```suggestion
      "! <p class="shorttext">JSON</p>
      "! JSON quick action
```

Markus1812 (changes requested):
Hi, thanks for contributing your object type. Please take a look at the comments below. If you agree with the suggestion, you can directly apply them here in GitHub and generate the new schema afterwards via your ABAP system.

Markus1812 (commented):
Hi, sorry for the delay. While reviewing again, I have some further questions.

LutzR-SAP:
> @Markus1812
> 
> > `How are these JSONs created? Are they generated by the system or do you plan to include some actions that allow the user to edit this field?`
> 
> At the moment (SAP Developer Use-Case) they are written in JSON-Editor and uploaded. We are currently working on an Agent to generate these JSONs. Yes, we plan to have this JSON editable.
> 
> > `My biggest concern is that although the AFF can be used for transporting objects, it also has the goal to be understandable and editable by users`
> 
> Yes, I agree. I was thinking only only of the Form-Based Editor so far. Just an idea: Is it possible to have a seperate File for this JSON-content (similar to Class source code)?

Hi @Markus1812, @schneidermic0,
How can we proceed here?
We can also setup a meeting, if that helps.

Best regards,
Raffael

schneidermic0:
> How can we proceed here?
We can also setup a meeting, if that helps.

Yes, I believe a meeting to understand your object type in more detail would be helpful.

GuilhermeSaraiva96:
Last message in this PR was over a month ago. Sorry if I may be out of the topic, but are there any updates?

LutzR-SAP:
> Last message in this PR was over a month ago. Sorry if I may be out of the topic, but are there any updates?

Hi @GuilhermeSaraiva96,
Unfortunately, we had to deprioritize this topic temporarily, to focus on more urgent deliverables.
However, we want to continue on this. So I will set up a meeting

schneidermic0:
We had a short meeting today. The idea is to move the dedicated JSON data in a separate file. The format of the other JSON does already exist in review booklets and support features not covered by standard AFF JSONs.

@LutzR-SAP Object types (like CSNM or SFPF) support already multiple files per object.

LutzR-SAP:
> We had a short meeting today. The idea is to move the dedicated JSON data in a separate file. The format of the other JSON does already exist in review booklets and support features not covered by standard AFF JSONs.
> 
> @LutzR-SAP Object types (like CSNM or SFPF) support already multiple files per object.

Hi @schneidermic0,
I adjusted the file format as discussed and introduced "<name>.abiq.content.json" for the JSON-content.
The step "[Validate examples against schema](https://github.com/SAP/abap-file-formats/actions/runs/28011322381/job/82904938182?pr=745#logs)" is now failing. Is this due to the newly introduced JSON-File?

LutzR-SAP:
During testing I realized, that we have a similar situation for the UserPrompt as well.
It's just a text-field, but it contains line-breaks and some formatting. So it is reather hard to read/write that from the aff-File tself.
My suggestion is there to extract that file in a separate file (*.prompt.txt) as well.

Feedback is of course welcome!

schneidermic0:
> > We had a short meeting today. The idea is to move the dedicated JSON data in a separate file. The format of the other JSON does already exist in review booklets and support features not covered by standard AFF JSONs.
> > @LutzR-SAP Object types (like CSNM or SFPF) support already multiple files per object.
> 
> Hi @schneidermic0, I adjusted the file format as discussed and introduced ".abiq.content.json" for the JSON-content. The step "[Validate examples against schema](https://github.com/SAP/abap-file-formats/actions/runs/28011322381/job/82904938182?pr=745#logs)" is now failing. Is this due to the newly introduced JSON-File?

Sorry for the late response. Yes we need to create an exemption for this. I'll prepare a PR for this.

schneidermic0:
> During testing I realized, that we have a similar situation for the UserPrompt as well. It's just a text-field, but it contains line-breaks and some formatting. So it is reather hard to read/write that from the aff-File tself. My suggestion is there to extract that file in a separate file (*.prompt.txt) as well.
> 
> Feedback is of course welcome!

I see no better solution for this for the file formats as of now.

schneidermic0:
> > > We had a short meeting today. The idea is to move the dedicated JSON data in a separate file. The format of the other JSON does already exist in review booklets and support features not covered by standard AFF JSONs.
> > > @LutzR-SAP Object types (like CSNM or SFPF) support already multiple files per object.
> > 
> > 
> > Hi @schneidermic0, I adjusted the file format as discussed and introduced ".abiq.content.json" for the JSON-content. The step "[Validate examples against schema](https://github.com/SAP/abap-file-formats/actions/runs/28011322381/job/82904938182?pr=745#logs)" is now failing. Is this due to the newly introduced JSON-File?
> 
> Sorry for the late response. Yes we need to create an exemption for this. I'll prepare a PR for this.

@LutzR-SAP I am about to prepare the PR. I was wondering whether you have actually a schema that could be shared in this repository. If so, I could maybe link the schema to our validation

schneidermic0:
#808 ignores the file, but will bring the changes required to solve the validation issue in this PR

LutzR-SAP:
> @LutzR-SAP I am about to prepare the PR. I was wondering whether you have actually a schema that could be shared in this repository. If so, I could maybe link the schema to our validation

@schneidermic0  I added the schema [abiq-v1-content.json](https://github.com/LutzR-SAP/abap-file-formats/blob/aibq/new-aff/file-formats/abiq/abiq-v1-content.json)
Any suggestions regarding the name?

schneidermic0:
I tried to make it running in this PR. But there are more dependencies as of now that need to be considered. Therefore, I have removed the content file again and also ignored all example files that end with `abiq.content.json`.

I still like the idea of adding also the schema, but maybe we need to do this in a separate change. I will close #808

LutzR-SAP:
Hi @schneidermic0,
We would like like to stick with *.txt file for Prompts. While Markdown-Format is generally supported, we do not want to encourage complex instructions via the prompts.
The Idea of Analytical Quick-Actions is, to push down as much of the logic as possible to the backend (via JSON Quick-Actions) and only have simple textual instructions in a prompt.
Supporting *.md-Files might give users the impression, that complex instructions (akin to skills) are intended.
Therefore we think *.txt-Files are a better fit.

Are there any other open points, which prevent us from merging the PR and proceeding?

Best regards,
Raffael

schneidermic0:
> Are there any other open points, which prevent us from merging the PR and proceeding?

In general, I think the AFF is fine. For me, it is also fine to have a text file, we might want to add this in our documentation about supported file types.

I also found one small issue that you might want to fix. Your constants are based on `type n length 2`. However, the values are defined with 0, 1 and 2. This would work, but wouldn't it be cleaner to use the real numerics explicitly (i.e., '00', '01' and '02'?

schneidermic0:
> For me, it is also fine to have a text file, we might want to add this in our documentation about supported file types.

Done with https://github.com/SAP/abap-file-formats/pull/745/commits/0a2dbc917c79ec3b7fe1b5eec009dac34bbccb6b

PR #746, opened 2026-03-24
[CHKO] category and class must contain valid values

larshp:
don't allow empty values or lower case or strange characters


albertmink (approved):
LGTM

Issue #756, opened 2026-04-09
Machine readable meta format data


larshp:
As multiple object types are getting multiple files, I think we need some machine readable information to easily discover and check if files are permitted/possible.

This could be another top level JSON file, for each object type containing the list of versions along with the file cardinality eg like https://github.com/SAP/abap-file-formats/tree/main/file-formats/ddls#file-structure

PR #761, opened 2026-04-11
add meta format suggestion

larshp:
#756

to do later:
* add for each object type
* auto generate the README.md files


file-formats/clas/clas.json
```diff
@@ -0,0 +1,53 @@
...
+  "versions": [
+    {
+      "version": 1,
+      "files": [
+        {
+          "filePattern": "<name>.clas.json",
+          "cardinality": "1",
+          "schema": "clas-v1.json",
+          "definitionInterface": "zif_aff_clas_v1",
+          "examples": ["examples/z_aff_example_clas.clas.json"]
```
schneidermic0: Do we need this information? If so, shouldn't we put also a relative linke to the ABAP source for the definition interface?

file-formats/clas/clas.json
```diff
@@ -0,0 +1,53 @@
+{
+  "$schema": "../meta-schema.json",
+  "objectType": {
+    "name": "CLAS",
+    "description": "Class"
+  },
+  "versions": [
+    {
+      "version": 1,
```
schneidermic0: format version is a `string`. Should we use a string here, too?

schneidermic0 (commented):
Thanks for the proposal: I have added some small questions.

file-formats/clas/clas.json
larshp: updated

file-formats/clas/clas.json
larshp: I think we will eventually auto generate the README.md

updated

schneidermic0 (approved):
LGTM for a first version. We could start with this.

larshp:
@albertmink any thoughts?

albertmink (commented):
@larshp be it a README file or a JSON Schema, both is fine. As we don't touch them that often, I guess keeping them in sync is a minor concern.

1) How do we create that file for all the object types?

2) Would you mind showing an example of that JSON file for an fictive object type that has two versions?
I'm not sure if I got the versioning right.

larshp:
1: I was thinking to ask the LLM to do it
2: each version is another entry in this array, https://github.com/SAP/abap-file-formats/pull/761/changes#diff-5bc7686ca8e29da2d57cf25a4eb59dc087d514af6d6e0cae28a73a3575965f77R7-R9

albertmink:
Hi @larshp 

about the versioning. How is the version, e.g. `2` linked to the files that correspond to this specific version? The array order is not fix I recon.

larshp:
the files and formats/schemas are defined for each version,  

```json
  "versions": [
    {
      "version": "1",
      "files": [something]
    },
    {
      "version": "1",
      "files": [something new]
    }
```

you mean we get a meta version number on top of the existing?

albertmink:
Ohh, got it. It is an object {...} that holds all information wrt a version. Never mind, it's alright.

albertmink (approved):
LGTM

albertmink (approved):
Let'd do this and have this JSON Schema.

schneidermic0:
Thanks @albertmink and @larshp for bringing this forward. 

Sorry, I am not fast enough and get distracted by other topics, currently. :(

PR #737, opened 2026-02-03
AFF for Lock Objects (ENQU)

anne-hobbs:
AFF definition for Lock Objects (ENQU)

PR #768, opened 2026-04-20
Fix ABAP Doc Casing Check permissions for fork PRs

wurzka:
## Problem

The ABAP Doc Casing Check workflow fails on PRs from forks with the error:
```
GraphQL: Resource not accessible by integration (addComment)
```

This is visible in PR #737 where the "Post casing checklist on PR" job fails.

## Root Cause

When using `pull_request` trigger, workflows from fork PRs run with read-only `GITHUB_TOKEN` permissions as a security measure. Even though we declare `pull-requests: write` in permissions, GitHub restricts this for fork PRs.

## Solution

Changed from `pull_request` to `pull_request_target`:
- `pull_request_target` runs in the context of the base repository (not the fork)
- This gives the workflow write permissions to post comments
- Safe for this workflow since it only posts a checklist comment, no code execution from the fork

Also updated event references from `github.event.pull_request.number` to `github.event.number` as the PR number is at the top level for `pull_request_target` events.

## Testing

This fix will allow the workflow to successfully post checklist comments on fork PRs like #737.


albertmink (approved):
LGTM. Maybe this Action is getting replace with the Hyperspace bot 🤖

PR #770, opened 2026-04-24
[DYNP] first draft

larshp:
with abapGit as starting point, probably nobody knows the real format, so we need to start somewhere


larshp:
@albertmink @schneidermic0 something like this could be a starting point, so we can move e.g. PROG and FUGR in abapGit to AFF?

schneidermic0:
I have reached out to some colleagues whether they can provide input for dynpros.

cohomology:
Dynpro Team has it as official backlog item

larshp:
see #819 instead

gonna keep this PR open for a few more weeks, then close

PR #782, opened 2026-05-12
Adapt description length for FUGR objects

Christopher-Hermann:
FUGR description length = 40
FUNC description length = 74
REPS description length = 70


larshp:
`zif_aff_types_v1=>ty_header_only_description` can be removed with this change, or alternatively adjust it?

file-formats/fugr/type/zif_aff_fugr_v1.intf.abap
```diff
@@ -28,6 +28,21 @@ INTERFACE zif_aff_fugr_v1
...
     END OF co_status.
 
+  "! <p class="shorttext">Description</p>
+  "! Description of the ABAP object
+  TYPES ty_description_40 TYPE c LENGTH 40.
+
+  TYPES:
+    "! <p class="shorttext">Header</p>
+    "! The header for an ABAP main object (with source code) with a description of 40 characters
+    BEGIN OF ty_header_40_src,
```
schneidermic0: It's really a minor, but we could remove all the additions we had in `zif_aff_types_v1` so that it is only `ty_description` and `ty_header`. What do you think? 

Same in the other files.

schneidermic0 (dismissed):
In general, I think this is good. I had only a minor suggestion and I agree with @larshp. If we don't need `zif_aff_types_v1 =>ty_header_only_description` we could delete it.

If no one has objections I suggest to do the incompatible change here to solve the bug. Anyhow, there must not be any object with longer descriptions. We can consider this as a bug. Any objections?

file-formats/fugr/type/zif_aff_fugr_v1.intf.abap
Markus1812: 👍

Christopher-Hermann:
Thanks @Markus1812 for taking over 👍

wurzka (approved):
LGTM 👍

PR #783, opened 2026-05-19
Change ENHO AFF

huber-nicolas:
## Summary

- Restructures `enhancementSpot` into a `generalInformation` object in `enho-v1.json`
- Expands BAdI implementation schema with additional filter value types and properties
- Updates `zif_aff_enho_v1.intf.abap` interface definitions to match the new schema
- Updates example file `z_aff_example_enho.enho.json` to reflect the new structure

## Changed Files

- `file-formats/enho/enho-v1.json` — JSON schema updated
- `file-formats/enho/examples/z_aff_example_enho.enho.json` — example updated
- `file-formats/enho/type/zif_aff_enho_v1.intf.abap` — ABAP interface updated

## Test plan

- [ ] Validate `enho-v1.json` schema is well-formed
- [ ] Verify example file conforms to the updated schema
- [ ] Check ABAP interface aligns with the schema changes


file-formats/enho/type/zif_aff_enho_v1.intf.abap
```diff
@@ -63,111 +31,178 @@ INTERFACE zif_aff_enho_v1
...
-      "! where "*" represents any character string (including a blank string) and "+" represents any character.
       "! It is not case-sensitive.
       covers_pattern     TYPE ty_comparator VALUE 'CP',
       "! <p class="shorttext">No Pattern</p>
-      "! No Pattern: True, if a logical expression with CP is false, that is, if operand1 does not fit the pattern
-      "! operand2. If the comparison is false, sy-fdpos contains the offset of operand2 in operand1, whereby
-      "! leading wildcard characters "*" in operand2 are ignored if operand2 also contains other characters.
-      "! If the comparison is true, sy-fdpos contains the length of operand1.
+      "! No Pattern: True, if a logical expression with CP is false, that is, if operand1 does not fit the pattern operand2.
       no_pattern         TYPE ty_comparator VALUE 'NP',
```
huber-nicolas:
```suggestion
      "! No Pattern: True, if a logical expression with CP is false, that is, if operand1 does not fit the 
      "! pattern operand2.
no_pattern         TYPE ty_comparator VALUE 'NP',
```

file-formats/enho/type/zif_aff_enho_v1.intf.abap
```diff
@@ -63,111 +31,179 @@ INTERFACE zif_aff_enho_v1
...
       covers_pattern     TYPE ty_comparator VALUE 'CP',
       "! <p class="shorttext">No Pattern</p>
-      "! No Pattern: True, if a logical expression with CP is false, that is, if operand1 does not fit the pattern
-      "! operand2. If the comparison is false, sy-fdpos contains the offset of operand2 in operand1, whereby
-      "! leading wildcard characters "*" in operand2 are ignored if operand2 also contains other characters.
-      "! If the comparison is true, sy-fdpos contains the length of operand1.
-      no_pattern         TYPE ty_comparator VALUE 'NP',
+      "! No Pattern: True, if a logical expression with CP is false, that is, if operand1 does not fit the 
+      "! pattern operand2.
+no_pattern         TYPE ty_comparator VALUE 'NP',
```
huber-nicolas:
```suggestion
      no_pattern         TYPE ty_comparator VALUE 'NP',
```

file-formats/enho/type/zif_aff_enho_v1.intf.abap
```diff
@@ -63,111 +31,179 @@ INTERFACE zif_aff_enho_v1
...
-      "! Comparators
-      comparators        TYPE ty_filter_comparators,
-      "! <p class="shorttext">Filter Values</p>
-      "! Filter values
-      values             TYPE ty_filter_values,
-      "! <p class="shorttext">Property Reference</p>
-      "! Property reference
-      property_reference TYPE string,
-    END OF ty_badi_impl_filter.
+      filtertype TYPE ty_filtertype, " read only / ENHS
```
huber-nicolas:
```suggestion
      filtertype TYPE ty_filtertype,
```

file-formats/enho/examples/z_aff_example_enho.enho.json
marcushoepfner, line 2: maybe a stupid question, but do we need to increase `formatVersion` then?

file-formats/enho/type/zif_aff_enho_v1.intf.abap
```diff
@@ -88,86 +56,158 @@ INTERFACE zif_aff_enho_v1
...
+      and    TYPE STANDARD TABLE OF ty_level_2 WITH DEFAULT KEY,
+      "! <p class="shorttext">Filter</p>
+      "! Filter
+      filter TYPE ty_filter,
+    END OF ty_level_1.
+
+  TYPES:
+    "! <p class="shorttext">Level 0</p>
+    "! Level 0
+    BEGIN OF ty_level_0,
```
marcushoepfner: are there better names for these (as they appear in the schema with title and description)?
There are semantics behind these levels right? They are not all the same.

So we have or, and, or as possible levels.
If we call the outer or compination, like combination, and, or we could name the level like this.

huber-nicolas: "If we call the outer or compination, like combination, and, or we could name the level like this."
Sorry I don't understand this sentence :-)

huber-nicolas: can you please make a suggestion for the 3 levels?

file-formats/enho/examples/z_aff_example_enho.enho.json
huber-nicolas: no, we don't have to increase the version.

file-formats/enho/type/zif_aff_enho_v1.intf.abap
marcushoepfner: Sorry.
we have this (possible) hierarchy:
- outer or
- and
- inner or
and filter on every level of course.

If we call the outer or `combination` than we have:
- combination
- and
- or

Can that be reflected the in type names instead of `ty_level_0` and so on?

file-formats/enho/type/zif_aff_enho_v1.intf.abap
```diff
@@ -88,86 +56,158 @@ INTERFACE zif_aff_enho_v1
...
-      "! Comparator 2
-      comparator2 TYPE ty_comparator,
-    END OF ty_filter_comparators.
-
   "! <p class="shorttext">Filter Type</p>
   "! Filter type
-  "! $values {@link zif_aff_enho_v1.data:co_filter_type}
-  TYPES ty_filter_type TYPE c LENGTH 1.
+  "! $values {@link zif_aff_enho_v1.data:co_filtertype}
+  "! $default {@link zif_aff_enho_v1.data:co_filtertype.string}
```
schneidermic0: Did you check what is the most used type? I would also use either string or character-like as default.

file-formats/enho/type/zif_aff_enho_v1.intf.abap
```diff
@@ -1,42 +1,10 @@
...
-      supported              TYPE ty_customizing VALUE ' ',
-      "! <p class="shorttext">Customizing Supported (No Transport)</p>
-      "! Customizing is supported (without transport)
-      supported_no_transport              TYPE ty_customizing VALUE 'D',
-    END OF co_customizing.
-
   "! <p class="shorttext">Filter Comparator</p>
   "! Filter comparator
   "! $values {@link zif_aff_enho_v1.data:co_comparator}
+  "! $default {@link zif_aff_enho_v1.data:co_comparator.equal}
```
schneidermic0, line 7: I think this is a good idea to have equal as a default to. In the example you see that it (in this case) looks like a key-value pair for a filter. I like it. 👍

file-formats/enho/type/zif_aff_enho_v1.intf.abap
schneidermic0: This comment is also relevant for all other constants for the comparator. Would it make sense to render this value as "<" instead of "lessThan"? Would this increase the readability for you?

file-formats/enho/type/zif_aff_enho_v1.intf.abap
```diff
@@ -88,86 +56,158 @@ INTERFACE zif_aff_enho_v1
...
+      "! <p class="shorttext">BAdI Definition</p>
+      "! BAdI Definition of the BAdI implementation
+      "! $required
+      badi_definition           TYPE zif_aff_types_v1=>ty_object_name_30,
+      "! <p class="shorttext">Implementing Class</p>
+      "! Implementing class of the BAdI implementation
+      implementing_class        TYPE zif_aff_types_v1=>ty_object_name_30,
+      "! <p class="shorttext">Is Example Implementation</p>
+      "! BAdI implementation is an example implementation
+      "! $showAlways
```
schneidermic0: Do we really want to have these fields as "showAlways"?

file-formats/enho/examples/z_aff_example_enho.enho.json
schneidermic0: I am not sure. Officially we would have to increase the version. However, I believe this format version has not been used yet. So we could still change it incompatible.

file-formats/enho/type/zif_aff_enho_v1.intf.abap
```diff
@@ -88,86 +56,158 @@ INTERFACE zif_aff_enho_v1
...
+      "! BAdI implementation is active
+      "! $showAlways
+      is_active_implementation  TYPE abap_bool,
+      "! <p class="shorttext">BAdI Implementation Customizing</p>
+      "! Does the BAdI implementation support customizing
+      "! $showAlways
+      customizing               TYPE ty_customizing,
+      "! <p class="shorttext">Filter Values</p>
+      "! Filter values for this BAdI implementation
+      filter_values             TYPE ty_level_0,
```
schneidermic0: As far as I can see with this proposal we might get rid of "$callbackClass", don't we? 

If so, I suggest to remove also the logic from the generator in the tools-repository.

file-formats/enho/examples/z_aff_example_enho.enho.json
wurzka: Not sure whether you talked more about this. We might have to double check this again. Usually, we say that if the AFF is already contributed to abapGit, used by the XCO APIs or was delivered in an on premise release, we'll have to increase. I'm not sure whether we have to increase the format version because of the latter.

huber-nicolas: @albertmink ENHO was already delivered in an on premise release. Do you think we have to increase the version?

file-formats/enho/type/zif_aff_enho_v1.intf.abap
huber-nicolas: i have renamed the levels and have uploaded it

file-formats/enho/type/zif_aff_enho_v1.intf.abap
```diff
@@ -1,294 +1,296 @@
...
+      and    type tt_level_2,
+      "! <p class="shorttext">Filter</p>
+      "! Filter
+      filter type ty_filter,
+    end of ty_level_0.
+
+  "! <p class="shorttext">Customizing</p>
+  "! Customizing
+  "! $values {@link zif_aff_enho_v1.data:co_customizing}
+  "! $default {@link zif_aff_enho_v1.data:co_customizing.not_supported}
```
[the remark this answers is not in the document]
huber-nicolas: The rule says
*) Arrays don't need to be marked as $required in this case. In JSON it is possible to distinguish between empty arrays ("table": []) and arrays with "empty" array items ("table": [{}]). All fields with defaults in the array items that don't follow the other rules, along with their higher-level structures, must be marked as $required.

Since ty_customizing is directly inside a table I don't need a required for the top structures

file-formats/enho/examples/z_aff_example_enho.enho.json
```diff
@@ -15,9 +17,20 @@
       "isDefaultImplementation": true,
       "isActiveImplementation": true,
       "customizing": "notSupported",
-      "filterValues": [
-        "EXAMPLE_FILTER=example_value1"
-      ]
+      "filterValues": {
+        "or": [
```
marcushoepfner: Can this be changed. This is actually just one filter.
But we should also have a complex one with an and of or with two filters.

marcushoepfner (changes requested):

file-formats/enho/examples/z_aff_example_enho.enho.json
huber-nicolas: done in the newest commit

schneidermic0 (commented):
See some comments below. I am wondering whether the filter value ranges are covered with this proposal or not?

Sorry, I missed to post the review, yesterday.

file-formats/enho/enho-v1.json
```diff
@@ -134,69 +145,588 @@
...
+                                    "equal",
+                                    "lessThan",
+                                    "greaterThan",
+                                    "lessEqual",
+                                    "greaterEqual",
+                                    "notEqual",
+                                    "coversPattern",
+                                    "noPattern",
+                                    "containsString",
+                                    "containsNoString"
```
marcushoepfner: is this rather "noContainsString"?

file-formats/enho/enho-v1.json
```diff
@@ -134,69 +145,588 @@
...
+                                    "Numeric",
+                                    "Packed"
+                                  ],
+                                  "default": "string"
+                                },
+                                "comparator": {
+                                  "title": "Comparator",
+                                  "description": "Comparator",
+                                  "type": "string",
+                                  "enum": [
```
marcushoepfner, line 215: shall we rather go for
`=`, `<`, ...`<=`, `<>`, `*`, ...?
(only in `enum`, I would leave `enumTitles` as is)

file-formats/enho/type/zif_aff_enho_v1.intf.abap
huber-nicolas: yes, good catch. I will do so

file-formats/enho/type/zif_aff_enho_v1.intf.abap
huber-nicolas:
```
"filter": {
                  "filtername": "EXAMPLE_FILTER",
                  "comparator": "greaterThan",
                  "value": "1"
             },
versus
"filter": {
                  "filtername": "EXAMPLE_FILTER",
                  "comparator": ">",
                  "value": "1"
                }
```

I like the words more than the symbols. But I see that @marcushoepfner also likes the symbols more

file-formats/enho/enho-v1.json
huber-nicolas: same discussion as here
https://github.com/SAP/abap-file-formats/pull/783#discussion_r3271725235

file-formats/enho/type/zif_aff_enho_v1.intf.abap
huber-nicolas: I removed it

file-formats/enho/type/zif_aff_enho_v1.intf.abap
huber-nicolas: I have created an issue in the tools repo

file-formats/enho/enho-v1.json
wurzka: I don't have a strong feeling regarding `=` or `equals`. But what would it be for example for `CP`? Is there any symbol? I would prefer `containsPattern` to `CP`. But I'm against a mixture of all options

file-formats/enho/enho-v1.json
huber-nicolas: the names/descriptions of the comparators come mainly from here
https://help.sap.com/doc/abapdocu_cp_index_htm/CLOUD/en-US/ABENLOGEXP_STRINGS.html

file-formats/enho/enho-v1.json
marcushoepfner: I always have to translate the words into the mathematical representation in my brain.
When I read "=" it is clear for me, "equals" has to be translated into what it means.
Also if I think of a form-based UI where this might be a combo box, I like "=" and others a bit more.

file-formats/enho/type/zif_aff_enho_v1.intf.abap
huber-nicolas: I have analysed in our internal system. The frequency ranking of the filter types is:
character
string
numeric
integer
packed

I will set character as default

file-formats/enho/enho-v1.json
huber-nicolas: I have uploaded a new commit in which I am using the symbols

file-formats/enho/type/zif_aff_enho_v1.intf.abap
huber-nicolas: I am closing this comment. We can continue to discuss here
https://github.com/SAP/abap-file-formats/pull/783#discussion_r3278948430

file-formats/enho/enho-v1.json
huber-nicolas: In the newest commit I have updated all comparators and use their official abapdocu as title/descriptions

huber-nicolas:
The work on this pull request is paused due to vacation season

file-formats/enho/type/zif_aff_enho_v1.intf.abap
```diff
@@ -1,173 +1,229 @@
...
-      property_reference TYPE string,
-    END OF ty_badi_impl_filter.
+      filtertype TYPE ty_filtertype,
+      "! <p class="shorttext">Comparator</p>
+      "! Comparator
+      comparator TYPE ty_comparator,
+      "! <p class="shorttext">Filter Value</p>
+      "! Filter value
+      value      TYPE string,
+    END OF ty_filter.
```
wurzka, line 108: There is some duplicated `filter`. What about the following:

```suggestion
    BEGIN OF ty_filter,
"! <p class="shorttext">Name</p>
"! Filter name
      name TYPE zif_aff_types_v1=>ty_object_name_30,
"! <p class="shorttext">Type</p>
"! Filter type
      type TYPE ty_filtertype,
      "! <p class="shorttext">Comparator</p>
      "! Comparator
      comparator TYPE ty_comparator,
      "! <p class="shorttext">Filter Value</p>
      "! Filter value
      value      TYPE string,
    END OF ty_filter.
```

file-formats/enho/type/zif_aff_enho_v1.intf.abap
```diff
@@ -212,60 +268,27 @@ INTERFACE zif_aff_enho_v1
...
-    END OF ty_badi_impl,
 
-    "! <p class="shorttext">BAdI Implementations of the ENHO</p>
-    "! BAdI implementations of the ENHO
-    ty_badi_impls TYPE STANDARD TABLE OF ty_badi_impl WITH DEFAULT KEY.
+  TYPES:
+    "! <p class="shorttext">General Information</p>
+    "! General information"
+    BEGIN OF ty_general_information,
+      "! <p class="shorttext">Enhancement Spot Name</p>
```
wurzka:
```suggestion
      "! <p class="shorttext">Enhancement Spot</p>
```
Would be ok for me in the description

file-formats/enho/type/zif_aff_enho_v1.intf.abap
```diff
@@ -278,13 +301,13 @@ INTERFACE zif_aff_enho_v1
       "! Header
       "! $required
       header               TYPE zif_aff_types_v1=>ty_header_100,
-      "! <p class="shorttext">Enhancement Spot Name</p>
-      "! Enhancement Spot name
+      "! <p class="shorttext">General Information</p>
+      "! General Information
```
wurzka:
```suggestion
      "! General information
```

file-formats/enho/enho-v1.json
```diff
@@ -1,3 +1,4 @@
+
```
wurzka: This line is not intended, is it?

wurzka (changes requested):
See my comments

file-formats/enho/type/zif_aff_enho_v1.intf.abap
```diff
@@ -1,173 +1,229 @@
...
+      "! $enumValue 'CS'
+      contains_string             TYPE ty_comparator VALUE 'CS',
       "! <p class="shorttext">Contains No String</p>
-      "! Contains No String: True, if a logical expression with CS is false, that is if operand1 does not contain the
-      "! content of operand2. If the comparison is false, sy-fdpos contains the offset of operand2 in operand1.
-      "! If the comparison is true, sy-fdpos contains the length of operand1.
-      contains_no_string TYPE ty_comparator VALUE 'NS',
+      "! True, if a comparison expression with CS is false, that is if operand1 does not contain the content
+      "! of operand2.
+      "! If the comparison is false, sy-fdpos contains the offset of operand2 in operand1. If the comparison
```
marcushoepfner: ...sy-fdpos... strange description isn't it?
I don't think we need such a description on meta model level. This is rather what happens at runtime.

file-formats/enho/type/zif_aff_enho_v1.intf.abap
```diff
@@ -212,60 +268,27 @@ INTERFACE zif_aff_enho_v1
...
       "! Main object name of the referenced object
       "! $required
-      main_object_name TYPE  c LENGTH 40,
+      main_object_name TYPE c LENGTH 40,
     END OF ty_referenced_object,
 
     "! <p class="shorttext">Referenced Objects</p>
     "! Referenced objects of the enhancement implementation
-    ty_referenced_objects TYPE STANDARD TABLE OF ty_referenced_object WITH DEFAULT KEY,
+    ty_referenced_objects TYPE STANDARD TABLE OF ty_referenced_object WITH DEFAULT KEY.
```
marcushoepfner: not sure whether we need this information. seems to be a runtime information only.

marcushoepfner (changes requested):

file-formats/enho/type/zif_aff_enho_v1.intf.abap
huber-nicolas: you are right, this sy-fdpos does not make sense in the AFF use cases. I will remove it (also in the descriptions above)

file-formats/enho/type/zif_aff_enho_v1.intf.abap
huber-nicolas: done

file-formats/enho/type/zif_aff_enho_v1.intf.abap
huber-nicolas: I think you are right. I will remove it

file-formats/enho/type/zif_aff_enho_v1.intf.abap
```diff
@@ -1,231 +1,177 @@
...
+      "! <p class="shorttext">Customizing Not Supported</p>
+      "! Customizing is not supported
+      not_supported          TYPE ty_customizing VALUE 'X',
+      "! <p class="shorttext">Customizing Supported</p>
+      "! Customizing is supported
+      supported              TYPE ty_customizing VALUE ' ',
+      "! <p class="shorttext">Customizing Supported (No Transport)</p>
+      "! Customizing is supported (without transport)
+      supported_no_transport TYPE ty_customizing VALUE 'D',
+    END OF co_customizing.
```
wurzka, line 168: Since we usually recommend having title and component name in sync, we should do it here as well
```suggestion
    BEGIN OF co_customizing,
      "! <p class="shorttext">Not Supported</p>
      "! Customizing is not supported
      not_supported          TYPE ty_customizing VALUE 'X',
      "! <p class="shorttext">Supported</p>
      "! Customizing is supported
      supported              TYPE ty_customizing VALUE ' ',
      "! <p class="shorttext">Supported (No Transport)</p>
      "! Customizing is supported (without transport)
      supported_no_transport TYPE ty_customizing VALUE 'D',
    END OF co_customizing.
```

wurzka (changes requested):
Thanks for the changes @huber-nicolas.
I found one mire issue regarding titles and components that are not in sync. Maybe it make sense to check the whole AFF again (or let it check 😉 ). Since ENHO was one of our first file formats, there might be more names/titles that have to be adapted.

file-formats/enho/type/zif_aff_enho_v1.intf.abap
huber-nicolas: i have checked the whole AFF and adapted titles and field names in my newest commit

file-formats/enho/examples/z_aff_example_enho.enho.json
```diff
@@ -4,46 +4,41 @@
...
-    },
-    {
-      "objectType": "INTF",
-      "objectName": "Z_AFF_EXAMPLE_BADI_INTF",
-      "programId": "R3TR",
-      "elementUsage": "usedObject",
-      "mainObjectType": "INTF",
-      "mainObjectName": "Z_AFF_EXAMPLE_BADI_INTF"
+      "filterValues": {
+        "or": [
```
marcushoepfner: the `or` is weird, isn't it.
it contains and item (`and`) only.

marcushoepfner (changes requested):
found one more issue

file-formats/enho/examples/z_aff_example_enho.enho.json
wurzka: But that's the fault of the example, isn't it? The AFF would allow an `and` at first level.

huber-nicolas: i will correct it. I have written the example manually

wurzka (approved):
Thanks, looks good to me, now 👍

marcushoepfner (approved):

file-formats/enho/examples/z_aff_example_enho.enho.json
huber-nicolas: We will keep the format version 1

PR #784, opened 2026-05-29
revised aff: HTTP

I748585:
Dear colleagues,

although this AFF has previously (11 months ago) passed the AFF review, it was never pursued any further due to open questions regarding the UI. After clarification of those open questions it became clear that the interface IF_AFF_HTTP_V1 needs to be changed.

Mostly, the field "URL" that was removed during the previous review was re-added. However, now it is intended to only show the fixed, relative part of the URL without the host/port that needs to be calculated for each system/appserver.

As it was never released, we decided to change V1 instead of creating V2.


I748585:
The compatibility check fails with 'Error: Something was added'. Is that because one is not allowed to update AFFs once approved?

Please see initial comments for the reasoning behind this change (vs. a new version V2)

albertmink:
Hi @I748585
no worries. The compatibility check is not required, so not blocking.

I748585:
Thanks for the reply, Albert! Btw, the PR is ready for review. Unfortunately, I don't have the necessary authorizations to set the label 'awaiting-review' myself as described in the process. Not sure if I'm stating the obvious here, but better safe than sorry.

Best regards, Manuel

wurzka (approved):
Hi @I748585,
I'm sorry for the late response. File format wise it looks good to me.
I just added the issue #790 so that we don't forget to add an example here

wurzka:
Sorry, I forgot to merge. Thanks for updating the AFF 👍 
As already mentioned, increasing the format version is not necessary.

PR #794, opened 2026-06-16
[ENHO] Replace flat filter fields with normalFilterDetails and rangeFilterDetails

huber-nicolas:
## Summary

- Remove the `type` field from the filter object (filter type is read directly from the BAdI definition on the ABAP side, not stored in the AFF file)
- Replace the flat `comparator` + `value` fields with a `normalFilterDetails` sub-object containing `comparator` and `value`
- Add `isRangeFilter` boolean flag and `rangeFilterDetails` sub-object (`lowComparator`, `lowValue`, `highComparator`, `highValue`) for range filters
- Update the example file to demonstrate both a normal filter and a range filter

## Motivation

The filter type is defined on the BAdI definition (enhancement spot) and does not need to be duplicated in the AFF file. Separating normal and range filter details makes the structure clearer and avoids ambiguity between single-value and range filter comparators.

## Test plan

- [ ] Verify JSON schema is valid
- [ ] Verify example file validates against the updated schema
- [ ] Verify round-trip serialize/deserialize works with the updated ABAP implementation


huber-nicolas:
Closing in favour of a clean PR based on current main (no merge conflicts)

PR #795, opened 2026-06-16
[ENHO] Remove filtertype, rename low to left, rename high to right, add range structure

huber-nicolas:
## Summary

- Remove the `type` field from the filter object (filter type is read directly from the BAdI definition on the ABAP side, not stored in the AFF file)
- Replace the flat `comparator` + `value` fields with a `normalFilterDetails` sub-object containing `comparator` and `value`
- Add `isRangeFilter` boolean flag and `rangeFilterDetails` sub-object (`lowComparator`, `lowValue`, `highComparator`, `highValue`) for range filters
- Update the example file to demonstrate both a normal filter and a range filter

## Motivation

The filter type is defined on the BAdI definition (enhancement spot) and does not need to be duplicated in the AFF file. Separating normal and range filter details makes the structure clearer and avoids ambiguity between single-value and range filter comparators.

## Test plan

- [ ] Verify JSON schema is valid
- [ ] Verify example file validates against the updated schema
- [ ] Verify round-trip serialize/deserialize works with the updated ABAP implementation


file-formats/enho/examples/z_aff_example_enho.enho.json
```diff
@@ -10,27 +10,32 @@
...
           {
             "filter": {
-              "name": "EXAMPLE_STRING_FILTER",
-              "type": "string",
-              "value": "TEST"
+              "name": "EXAMPLE_RANGE_FILTER",
+              "isRangeFilter": true,
+              "rangeFilterDetails": {
+                "lowValue": "10",
+                "highComparator": "<=",
```
marcushoepfner: lowComparator is missing

file-formats/enho/type/zif_aff_enho_v1.intf.abap
```diff
@@ -62,49 +82,52 @@ INTERFACE zif_aff_enho_v1
...
-      "! Filter type
-      type       TYPE ty_filtertype,
-      "! <p class="shorttext">Comparator</p>
-      "! Comparator
-      comparator TYPE ty_comparator,
-      "! <p class="shorttext">Filter Value</p>
-      "! Filter value
-      value      TYPE string,
+      name                  TYPE zif_aff_types_v1=>ty_object_name_30,
+      "! <p class="shorttext">Normal Filter</p>
```
marcushoepfner: `normal` is a hard word.
just `filter`?

file-formats/enho/type/zif_aff_enho_v1.intf.abap
```diff
@@ -62,49 +82,52 @@ INTERFACE zif_aff_enho_v1
...
+      "! Normal filter
+      normal_filter_details TYPE ty_normal_filter,
+      "! <p class="shorttext">Is Range Filter</p>
+      "! Is range filter
+      is_range_filter       TYPE abap_bool,
```
marcushoepfner: do we need this additional flag?
either (normal) filter or a rangeFilter is filled.

file-formats/enho/type/zif_aff_enho_v1.intf.abap
```diff
@@ -1,6 +1,26 @@
...
+  "! <p class="shorttext">Range Filter Comparator</p>
+  "! Range filter comparator
+  "! $values {@link zif_aff_enho_v1.data:co_range_filter_comparator}
+  "! $default {@link zif_aff_enho_v1.data:co_range_filter_comparator.less_than}
+  TYPES ty_range_filter_comparator TYPE c LENGTH 2.
+
+  CONSTANTS:
+    "! <p class="shorttext">Range Filter Comparator</p>
+    "! Range filter comparator
+    BEGIN OF co_range_filter_comparator,
```
marcushoepfner, line 13: is this really just only `<` and `<=`?

I see:

Image

marcushoepfner (changes requested):

file-formats/enho/type/zif_aff_enho_v1.intf.abap
huber-nicolas: We do not need it. I remove it

file-formats/enho/type/zif_aff_enho_v1.intf.abap
huber-nicolas: Yes, I will rename it to just "filter"

file-formats/enho/examples/z_aff_example_enho.enho.json
huber-nicolas: it is not serialized, because it is the default. Enums need to be either required or have a default

file-formats/enho/type/zif_aff_enho_v1.intf.abap
huber-nicolas: In ADT it is only < and <=
image

file-formats/enho/examples/z_aff_example_enho.enho.json
marcushoepfner: oh, that's bad isn't it. end user would always have to know the default to understand.

huber-nicolas: i made it required

marcushoepfner (dismissed):

file-formats/enho/examples/z_aff_example_enho.enho.json
```diff
@@ -10,27 +10,32 @@
...
           },
           {
             "filter": {
-              "name": "EXAMPLE_STRING_FILTER",
-              "type": "string",
-              "value": "TEST"
+              "name": "EXAMPLE_RANGE_FILTER",
+              "rangeFilterDetails": {
+                "lowValue": "10",
+                "lowComparator": "<",
```
schneidermic0: Minor: I believe the order is wrong, isn't it?
```suggestion
                "lowComparator": "<",
                "lowValue": "10",

```

huber-nicolas: yes, you are right. The ST would serialize the example with the correct order, but I have created the example by hand

file-formats/enho/examples/z_aff_example_enho.enho.json
```diff
@@ -10,27 +10,32 @@
...
-        "and": [
+        "or": [
           {
             "filter": {
-              "name": "EXAMPLE_INTEGER_FILTER",
-              "type": "integer",
-              "comparator": ">",
-              "value": "1"
+              "name": "EXAMPLE_NORMAL_FILTER",
+              "filterDetails": {
```
marcushoepfner: i wonder if we should skip `filterdetails` as this is the standardfilter and range filters occur only occassionally.
we had one more level of nesting of many objects just because of the existence of a range filter. (and with or and and there already deep nesting)


my proposal for standard filter:
```json
"filter": {
              "name": "EXAMPLE_FILTER",
              "comparator": "=",
              "value": 42
            }
```

and range filter:
```json
"filter": {
              "name": "EXAMPLE_RANGE_FILTER",
              "range": {
                "lowComparator": "<",
                "lowValue": "10",
                "highComparator": "<=",
                "highValue": "100"
              }
```

With that we loose the ability to make comparator and value required. But for me this was ok.
WDYT?

huber-nicolas: yes, I like your idea. I will upload another commit later

file-formats/enho/examples/z_aff_example_enho.enho.json
```diff
@@ -10,27 +10,32 @@
...
             }
           },
           {
             "filter": {
-              "name": "EXAMPLE_STRING_FILTER",
-              "type": "string",
-              "value": "TEST"
+              "name": "EXAMPLE_RANGE_FILTER",
+              "rangeFilterDetails": {
+                "lowComparator": "<",
```
marcushoepfner: we could also call this left and right instead of low and high?
just a proposal. up to you.

huber-nicolas: done in the new commit

file-formats/enho/examples/z_aff_example_enho.enho.json
huber-nicolas: i had to add a new default value "not_specified" as comparator value and put it as default. See my new commit

file-formats/enho/examples/z_aff_example_enho.enho.json
```diff
@@ -10,27 +10,30 @@
...
       "isActiveImplementation": true,
       "customizing": "notSupported",
       "filterValues": {
-        "and": [
+        "or": [
           {
             "filter": {
-              "name": "EXAMPLE_INTEGER_FILTER",
-              "type": "integer",
+              "name": "EXAMPLE_NORMAL_FILTER",
```
marcushoepfner:
```suggestion
              "name": "EXAMPLE_FILTER",
```

marcushoepfner (commented):
one minor.
and the title of the PR does not really fit anymore.
worth to change for the history?

marcushoepfner (approved):

PR #796, opened 2026-06-18
[DTEL] Update schema to category-based data type information

schragalx:


GuilhermeSaraiva96:
please fix the failing checks

uenal-akkaya:
@GuilhermeSaraiva96 Somehow the checks are not re-run. Can you check?

file-formats/dtel/examples/z_aff_example.dtel.json
```diff
@@ -1,20 +1,36 @@
...
   "fieldLabels": {
-    "short": "Example",
-    "shortLength": 7,
-    "medium": "Example field",
-    "mediumLength": 13,
-    "long": "Example data element field",
-    "longLength": 26,
-    "heading": "Example data element",
-    "headingLength": 20
+    "short": "",
```
GuilhermeSaraiva96: you could add the descriptions again, no?

file-formats/dtel/examples/z_aff_example.dtel.json
```diff
@@ -1,20 +1,36 @@
...
+    "shortLength": 10,
+    "medium": "",
+    "mediumLength": 20,
+    "long": "",
+    "longLength": 40,
+    "heading": "",
+    "headingLength": 55
+  },
+  "additionalProperties": {
+    "searchHelp": {
```
GuilhermeSaraiva96: weird that these fields are here, if they are empty. Do you want to put some dummy data into them? If they were serialised in vs code, then these fields would be gone ,since they are empty (= default). Same with all other empty fields in this example

file-formats/dtel/type/zif_aff_dtel_v1.intf.abap
```diff
@@ -24,56 +24,95 @@ INTERFACE zif_aff_dtel_v1 PUBLIC.
...
       "! $minimum 0
       "! $maximum 10
       short_length   TYPE i,
       "! <p class="shorttext">Medium</p>
       "! Medium field label
       medium         TYPE c LENGTH 20,
-      "! <p class="shorttext">Medium Length</p>
-      "! Medium length
+      "! <p class="shorttext">Length of Medium Field Label</p>
+      "! Note that the entered length is the upper limit for the lengths of all translations of the entered text.
```
GuilhermeSaraiva96: Is there really a recommended length? or was this AI written.

Additionally, this is too big for a description (In Eclipse this could even be shown as a tooltip I think).
To fully utilise this AFF description, the main point to accentuate is what the field represents as the title Length of Medium Field Label is a bit uninformative. I would say: "Maximum length of all translation of the medium field"

file-formats/dtel/type/zif_aff_dtel_v1.intf.abap
```diff
@@ -114,69 +151,57 @@ INTERFACE zif_aff_dtel_v1 PUBLIC.
     BEGIN OF ty_bidirectional_options,
       "! <p class="shorttext">Basic Direction</p>
       "! Basic direction
+      "! $showAlways
       basic_direction TYPE ty_basic_direction,
-
       "! <p class="shorttext">No Bidirectional Filtering</p>
       "! Deactivates bidirectional text filtering
       no_filtering    TYPE abap_bool,
```
GuilhermeSaraiva96, line 154: In the eclipse editor, the field is called "Bidi Filtering". Here it is the opposite. Do we want to harmonise this with the ADT Editor, or is this better?


FYI, in our Server Driven UI Editor / Json Aff Editor, the default for a boolean is false. Which means that when an abap_bool field is false, it will not be serialised. In your case, when there is bidi filtering, the field will not be serialised. As a rule of thumb, the field should represent a property that is more often false than true, which is why false is the default.

GuilhermeSaraiva96 (changes requested):

GuilhermeSaraiva96:
Hi @uenal-akkaya @schragalx ,
Regarding the checks:
- the schema check: you changed the file zif_aff_ddic_types_v1, which is also used by the doma object type.
- compatibility check: you can ignore it since the previous DTEL AFF was just a draft and was never used productively. The check makes sure that no incompatible changes are introduced, in order not to break older AFF versions.

file-formats/dtel/type/zif_aff_dtel_v1.intf.abap
uenal-akkaya: These texts are not AI written. We took it from the official documentation of the data elements for the field lengths SCRLEN_S, SCRLEN_M, SCRLEN_L and HEADLEN. We shortened and adjusted the text a little bit.

file-formats/dtel/examples/z_aff_example.dtel.json
uenal-akkaya: Done

file-formats/dtel/examples/z_aff_example.dtel.json
uenal-akkaya: Alex has filled in data. If these fields are initial they are not shown.

file-formats/dtel/examples/z_aff_example.dtel.json
uenal-akkaya: Descriptions added

uenal-akkaya:
> Hi @uenal-akkaya @schragalx , Regarding the checks:
> 
> * the schema check: you changed the file zif_aff_ddic_types_v1, which is also used by the doma object type.
> * compatibility check: you can ignore it since the previous DTEL AFF was just a draft and was never used productively. The check makes sure that no incompatible changes are introduced, in order not to break older AFF versions.

Is fixed with Alex's new change

file-formats/dtel/type/zif_aff_dtel_v1.intf.abap
uenal-akkaya: I have checked the description specification: https://github.com/SAP/abap-file-formats/blob/main/docs/json.md#description
A description must consist of a maximum of 255 characters.

The recommendation of lengths are from the official documentation in the data elements which is also delivered for customers.

Please review our proposal texts.
Here are our proposed description texts which we shortened and made more accentuated:
**shortLength** _(133 chars)_
The entered length is the upper limit for all translations. The field label maximum should be somewhat longer than the original text.
 
**mediumLength** _(168 chars)_
The entered length is the upper limit for all translations. The field label maximum should be somewhat longer than the original text. Recommended length: 15 characters.
 
**longLength** _(168 chars)_
The entered length is the upper limit for all translations. The field label maximum should be somewhat longer than the original text. Recommended length: 20 characters.
 
**headingLength** _(254 chars)_
The entered length is the upper limit for all translations. The field label maximum should be somewhat longer than the original text. Note that the heading is often displayed when editing list output, its length should not exceed the data element length.

file-formats/dtel/type/zif_aff_dtel_v1.intf.abap
GuilhermeSaraiva96: I think this is the only comment left answering from my review @uenal-akkaya

GuilhermeSaraiva96 (dismissed):
The changes look good to me. Just that 1 comment open from my side, please check it out, regarding the no_filtering nepvagitve, if it maybe could be reversed.

I added Michael so that he can approve it while I am on holidays

file-formats/dtel/type/zif_aff_dtel_v1.intf.abap
uenal-akkaya: The BiDi Filtering value is stored in table DD04L, field _bidictrlc_ (DD: No Filtering of BIDI Formatting Characters), which is 99.99% of the time _false_. The default is more often false than true and also the default value. Mostly it's then not shown in the Json Aff Editor.
Also in the XML-response in ADT the attribute is <dtel:deactivateBIDIFiltering>false</dtel:deactivateBIDIFiltering>, thus also mostly false. I don't know why in the  ADT Eclipse Editor it was decided to have it "Bidi Filtering" having the value shown flipped, positive formulating. I guess that it's better UX to have non-negating names.
But we had internally discussed the naming and decided that _no_filtering_ would be better and less confusing in the AFF / JSON editor. Also mostly this attribute won't we shown anyways.

file-formats/dtel/examples/z_aff_example.dtel.json
```diff
@@ -4,17 +4,33 @@
...
   },
-  "dataTypeInformation":{
-    "domainName": "Z_AFF_EXAMPLE_DOMA"
+  "dataTypeInformation": {
+    "category": "domain",
+    "typeName": "Z_AFF_EXAMPLE_DOMA",
+    "predefinedType": {
+      "dataType": "CHAR",
+      "length": 3
+    }
```
schneidermic0, line 12: Is the example correct? Isn't it that you should have either a domain name (that then describes the predefined type) or you use a predefined type?

Or does a data element really store redundantly the values of the domain?

schneidermic0 (commented):
See my comment in the example. Otherwise, it looks good to me.

file-formats/dtel/examples/z_aff_example.dtel.json
uenal-akkaya: When a data element is activated the dictionary checks the referenced domain and type information which is then persisted in the DDIC data base table. On the other hand, if a domain is activated then all dependent ddic objects such as DTEL will be updated regarding the type information. In that way the dictionary is consistent.
The information here in the AFF with predefinedType is not considered when having category "domain".
Example: You could also have written here in the dataType "CHAR" while having in the domain Z_AFF_EXAMPLE_DOMA as type "INT4" -> dictionary ignores the information of the DTEL predefinedType and considers the predefined type information of the Z_AFF_EXAMPLE_DOMA.

Another example we have checked was that when you have a data element DTEL1 which has category domain DOMA1 in system ABC with type char length 10 and the same domain DOMA1 in system XYZ but with type int8. In system ABC the DTEL1 has as type char length 10. When you now transport DTEL1 from ABC to XYZ then the DTEL1 has type int8 because at activation the domain type is considered in the current system of the DTEL.

For the AFF it depends if the predefinedType information is useful to show for category "domain" or not when having e.g. Code Review in git or in VSC editor? As a developer I would say that it's a nice to have information. But it would be confusing when importing a DTEL with reference to domain and then having in git another predefinedType than in the system where the DTEL was imported. But as in the example above, this can already happen in normal development where you have the same DTEL in different systems.
I am not sure if we can assume that a normal developer is aware of this circumstance.

I am indecisive if we should have the predefinedType for category domain persisted in the AFF. Both options are valid from my point of view. My gut feeling is that we should go for hiding the predefinedType, it can be shown optionally if used in an editor.

file-formats/dtel/README.md
```diff
@@ -12,4 +12,5 @@ DTEL  | Dictionary Data Elements | Dictionary
 
 File | Cardinality | Definition | Schema | Example
 :--- | :--- | :--- | :--- | :---
-`<name>.dtel.json` | 1 | [`zif_aff_dtel_v1.intf.abap`](./type/zif_aff_dtel_v1.intf.abap) | [`dtel-v1.json`](./dtel-v1.json) | [`z_aff_example.dtel.json`](./examples/z_aff_example.dtel.json)
+`<name>.dtel.json` | 1 | [`zif_aff_dtel_v1.intf.abap`](./type/zif_aff_dtel_v1.intf.abap) | [`dtel-v1.json`](./dtel-v1.json) | [`z_aff_example_domain.dtel.json`](./examples/z_aff_example_domain.dtel.json)
+`<name>.dtel.json` | 1 | [`zif_aff_dtel_v1.intf.abap`](./type/zif_aff_dtel_v1.intf.abap) | [`dtel-v1.json`](./dtel-v1.json) | [`z_aff_example_predefined_type.dtel.json`](./examples/z_aff_example_predefined_type.dtel.json)
```
schneidermic0: Thanks for adding a 2nd example. I would put both examples in one line in the last field and rename the column "Example**s**". 

If we keep two lines it looks like that DTELs require two files, but we have two examples for the one file.

I wanted to update the file with a proposal, but it seems I am not allowed to make changes. So I try to add it manually here:

```
File | Cardinality | Definition | Schema | Examples
`<name>.dtel.json` | 1 | [`zif_aff_dtel_v1.intf.abap`](./type/zif_aff_dtel_v1.intf.abap) | [`dtel-v1.json`](./dtel-v1.json) | [`z_aff_example_domain.dtel.json`](./examples/z_aff_example_domain.dtel.json) [`z_aff_example_predefined_type.dtel.json`](./examples/z_aff_example_predefined_type.dtel.json)
```

file-formats/dtel/examples/z_aff_example.dtel.json
schneidermic0: > My gut feeling is that we should go for hiding the predefinedType, it can be shown optionally if used in an editor.

Yes, I agree. This way @schragalx  updated also the examples, now

schneidermic0 (changes requested):
@schragalx The schema seems to be outdated. Would you update it, please?

The other checks we might have to overrule.

schragalx:
> @schragalx The schema seems to be outdated. Would you update it, please?
> 
> The other checks we might have to overrule.

updated schema with previous commit

schneidermic0 (approved):
As mentioned before, we need to overrule the compatibility check since this was not consumed, yet. 

Thanks @schragalx and welcome as a contributor 👍

PR #798, opened 2026-06-19
Idea for Introducing "oneOf"

schneidermic0:
**DISCLAIMER** This is just an idea. I share it to get some feedback here.

# Introduce `oneOf` (Union Types) Documentation for ABAP File Formats

### New Features

✨ Added documentation for a new `oneOf` union type feature in ABAP file formats, enabling developers to model mutually exclusive alternatives in JSON files using ABAP Doc annotations.

### Changes

* `docs/json.md`: Added a new section **"Union Types (oneOf)"** covering:
  - New ABAP Doc annotations: `$oneOfGroup`, `$oneOfDiscriminatorFor`, and `$oneOfValue` with supported syntax variants (plain names and `{@link ...}` references)
  - How discriminator values are resolved (explicit `$enumValue` or camel-case transformation)
  - Recommended usage guidelines to avoid duplication and ensure consistency with enum definitions
  - A minimal ABAP code example demonstrating a two-branch `payload_variant` group
  - Description of the generated JSON Schema output using `const` constraints and `required` fields per branch
  - **Deserialization behavior** clarifying that `oneOf` is enforced at the JSON Schema validation layer, not at the ABAP transformation layer
  - A note that changes to `oneOf` constraints are typically incompatible and require bumping `formatVersion`
  - Added the new section to the Table of Contents

- [ ] 🔄 Regenerate and Update Summary

<details>
<summary>PR Bot Information</summary>

**Version:** `1.26.0`

- Event Trigger: `issue_comment.edited`
- Correlation ID: `c09bcdb9-b7dd-4972-97c4-20d8feb9fd13`
</details>


wurzka (commented):
I like the approach 👍 
We have so many object types with components that are only required dependent on some boolean or enum value. 
Having a better validation option in the schema is an improvement.

marcushoepfner:
`"not": { "required": ["referencePayload"] }` could be added to the first branch,
while `"not": { "required": ["inlinePayload"] }` can be added to the second branch.

Otherwise it is not mutual exclusive and `referencePayload` and `inlinePayload` can co-exist in the same object.

Apart from that, looks really good, also with the `const` to always have exact match to a oneOf-branch.
Otherwise it might be hard to find the correct branch when not all properties are filled yet; the branch might not be uniquely identifiable.

PR #808, opened 2026-07-08
JSON Validator: Map `*.abiq.content.json` Files

schneidermic0:
ABIQ objects plan to have a dedicated JSON content file with a non-AFF schema. The file exists already (independent of AFF).


schneidermic0:
This is not sufficient as of now. The validation requires a `formatVersion` in the JSON and als the schema compare considers non-AFF-schemas. Therefore, I just ignore the specific content type in an exemption list.

PR #809, opened 2026-07-14
[PARA] New object type SET/GET Parameter

huber-nicolas:


file-formats/para/examples/z_aff_example_para.para.json
```diff
@@ -0,0 +1,10 @@
+{
+  "formatVersion": "1",
+  "header": {
+    "description": "Airline Code",
+    "originalLanguage": "en"
+  },
+  "generalInformation": {
+    "userSpecificValue": "LH"
+  }
```
schneidermic0: I think this is (as the name states) user-specific runtime data and does not belong to the AFF. I expect it is not part of the runtime object. I have double-checked it with the developer and he agreed


```suggestion
  }
```

schneidermic0 (changes requested):
@huber-nicolas I think we need to remove the runtime information in field "generalInforamtion.userSpecificValue". See below.

Otherwise, I think this looks good. The object type is quite trivial :)

file-formats/para/examples/z_aff_example_para.para.json
huber-nicolas: Thanks, I have removed general_information and user_specific_value

schneidermic0 (approved):
LGTM. Thanks @huber-nicolas

PR #820, opened 2026-07-30
SMTG: fix typos

larshp:


Markus1812:
Thanks for finding that! I've notified the object type owner, so that this is also updated in our next release. I would like to keep this open, until I get feedback from them.

Markus1812 (approved):
Just got feedback from the object type owners: fixed in ABAP, will be delivered with the next release.

PR #822, opened 2026-07-30
CHDO: fix typo

larshp:
maybe incompatible?


Markus1812:
> maybe incompatible?

Yes, most likely incompatible if the AFF is already in use. Changing the default would probably be compatible? But that's not the case here 😢

larshp:
ping?

PR #826, opened 2026-08-02
PROG: Set program type in include example

kennyhml:
The PROG schema defines the `programType` property to be one of
```json
"enum": [
  "executableProgram",
  "modulePool",
  "subroutinePool",
  "include"
],
```
With a default being defines as
```json
"default": "executableProgram"
```
The include example does not define a `programType` in its properties. By the default, it would then fall back to `executableProgram` which obviously makes no sense. I assume the default is set as intended and the include is simply missing the property.


GuilhermeSaraiva96 (approved):
Makes sense, well spotted!

PR #833, opened 2026-08-06
DTEL: add documentation support

larshp:
structured like [DOMA](https://github.com/SAP/abap-file-formats/tree/main/file-formats/doma)


larshp:
ping?

GuilhermeSaraiva96:
@uenal-akkaya Hi Ünal, please have a look at Lars' proposal

larshp:
@uenal-akkaya can you help? this should be a small change, its suggesting the same pattern that is already merged

uenal-akkaya:
Sorry, was on vacation. As example this is the same as for DOMA but none other object type, right?

larshp:
yea, its the same example, its currently only DOMA that also has it, but as we iterate through it I think more will need/have it

uenal-akkaya:
docu for DTEL is ok from my side

PR #834, opened 2026-08-06
meta format updates

larshp:


schneidermic0:
We were very recently looking into this, too. I would like to double-check with a colleague on this

PR #844, opened 2026-08-13
[SMBC] Update 2608

sepp4me:


schneidermic0:
@sepp4me I think our compatibility check currently just identifies whether new fields have been added. I/we need to double-check this. So far I could see you have only added new (non-mandatory/non-required) fields, right? Did I miss anything?

If so, I would consider it as compatible change following our rules https://github.com/SAP/abap-file-formats/blob/main/docs/json.md#format-versions-and-compatibility

PR #852, opened 2026-08-25
revised AFF: HTTP

I748585:
proposed changes include:

rename field 'URL' into 'Service Path'. More fitting name as the field only shows the relative path of the URL and not the dynamically constructed absolute URL (discussed in https://github.com/SAP/abap-file-formats/pull/784).

add field 'Authorization Default Values'. Previously planned to be included as a 'link'/'simple' action, adding a dedicated field in the UI has been proven as the preferred option as it is closer to the 'old' client-driven UI. The SUSH object is always created together with the HTTP Service object and is added to the same TR.

add field (boolean) 'Enable CSRF Token Handling'. Has been a requirement for a while, which significantly simplifies CSRF token handling for the application (which up until now has to call a function manually from a utility class that we offer). Switch to server-driven approach seems like the perfect opportunity to introduce the flag for us.


file-formats/http/type/zif_aff_http_v1.intf.abap
```diff
@@ -9,9 +9,15 @@ INTERFACE zif_aff_http_v1
...
       handler_class TYPE zif_aff_types_v1=>ty_object_name_30,
-      "! <p class="shorttext">URL</p>
-      "! URL
-      url           TYPE string,
+      "! <p class="shorttext">Service Path</p>
+      "! Service path
+      service_path  TYPE string,
+      "! <p class="shorttext">Authorization Default Values</p>
+      "! Configure authorization default values
+      auth_values   TYPE string,
```
GuilhermeSaraiva96: When a SUSH object has application type = 'HTTP', it will be associated to a specific HTTP. So that leads me to believe that the HTTP object comes before the SUSH.

When the SUSH is created, the HTTP service will be added to the SUSH's content. Does this mean that the content for the HTTP also needs to be updated, upon the SUSH's creation?

I can already see some redundancy here. The "connection" between the HTTP and SUSH needs to be centralised, cannot be saved in 2 different places. Either in the AFF for SUSH, or for in the HTTP. And I am betting it should rather be on the SUSH side.

GuilhermeSaraiva96 (commented):
I know you added the authValues for UI reasons, but the AFF is a data model and is completely decoupled from the UI. The UI representation is based on the AFF, not the other way around. And this is also why we ask the object type owners to first implement the AFF, and to not focus on the UI.

With this in mind, does the authValues actually fit the data model of HTTP?

file-formats/http/type/zif_aff_http_v1.intf.abap
I748585: The SUSH object is always created alongside the HTTP object and since the field does only include the (deterministic) name of the SUSH object, it does not (ever) need to be updated. But that of course also means that it is redundant information. As you stated above, the field was added for UI reasons. Is it possible to keep the field in the UI, including the forward-navigation, even if it is not included in the AFF? Or does it _have to_ be solved via a link action?

GuilhermeSaraiva96: Even if the field does not have to be updated, it does not mean that is redundant information. The tables in the backend make a link between the HTTP<->SUSH object types, but in order to not be redundant, this information should exist in only 1 place/table. The same for the AFFs, this information shall exist in the Aff, but only in 1 of them. I will take this one up to our team.

Regarding the UI issue: we have gotten a few requests in the past for UI fields that do not belong to the AFF. But this was rather for the form editor in Eclipse. Since the source view in vs code and eclipse corresponds to the AFF, I would find it incorrect to add the extra field, even if just to display, as it would then diverge from the abapgit data. It just would not be copy-paste able for example.

GuilhermeSaraiva96: @Markus1812 @wurzka what is your opinion?

wurzka: I would not add fields to the AFF that do not belong to the object, i.e. are not stored together with the other information of the object.

I748585: Alright, thanks for your input. I have removed it from the AFF (and also from the UI (and added a link-action instead)). I'll open a new PR with the updated AFF. Before I do so, is there any other points that need to be addressed?

GuilhermeSaraiva96: You do not need to create a new PR, a new change is enough. I guess this PR will only cover the token handling field then, and that one looks good to me.

I748585: I updated the PR, thanks for reviewing!

GuilhermeSaraiva96:
Summary of the change: Field url was removed, and token_handling + service path were added. I guess url->service_path is a rename

The renaming of url is an incompatible change, which means the format version needs to be increased. Is it implemented in abapgit? If not, then the aff has never been used and we can perform the change without increasing the format version

I748585:
> Summary of the change: Field url was removed, and token_handling + service path were added. I guess url->service_path is a rename
> 
> The renaming of url is an incompatible change, which means the format version needs to be increased. Is it implemented in abapgit? If not, then the aff has never been used and we can perform the change without increasing the format version

Correct, it is a rename to better represent what the field is actually showing after it has been decided not to show the absolute URL in https://github.com/SAP/abap-file-formats/pull/709 . The current AFF has not yet been used anywhere, so I don't think we need to increase the format version

GuilhermeSaraiva96:
There are 2 failing checks:

- the compatibility one: we can ignore this one as you've said the AFF is not used anywhere
- broken markdown link in the README.md file:

The AI analysis for the second check:
 The "Check links" failure in PR #852 is a broken link in file-formats/http/README.md:

  - The README links to: ./examples/http_test_service.http.json (plural examples)
  - The actual file was created at: file-formats/http/example/http_test_service.http.json (singular example)

  The fix is to either rename the directory from example → examples, or update the link in the README from
  ./examples/... → ./example/.... The convention in this repo is to use examples (plural) — check file-formats/clas/ or
  other object types — so the directory should be renamed to examples.

GuilhermeSaraiva96 (approved):
Looks good! Since this object type is not yet used by any application, the incompatible changes are ok

wurzka (approved):
Thanks for the updates, looks good to me.

PR #854, opened 2026-09-02
[SUSO] Add new object type SUSO

WDFdaniel:


WDFdaniel:
Notes:
The structure and information in AFF are mainly based on the native SUSO ADT editor to cover the information available there.
Object Class Description, Description of Authorization Fields (in ty_authorization_field) and Description of Activities (in ty_permitted_activity) will be read-only and be read from tables TOBCT, DD04T and TACTT.
The Activity Field Flag (is_activity_field) and the Access Category (access_category) are only relevant in cloud systems and have connection to SIA3 object (Authorization Object Extension). For Activity Field Flag and Access Category BADI implementation from SIA3 called in context of the get and save processes in current SUSO ADT editor implementation. Would that be ok?
Regarding the examples: The set of examples might be enhanced and renamed later on.

file-formats/suso/type/zif_aff_suso_v1.intf.abap
```diff
@@ -0,0 +1,213 @@
...
+      object_class             TYPE c LENGTH 4,
+      "! <p class="shorttext">Object Class Description</p>
+      "! Description of the authorization object class
+      object_class_description TYPE c LENGTH 60,
+    END OF ty_general_information.
+
+  TYPES:
+    "! <p class="shorttext">Authorization Fields</p>
+    "! Authorization fields
+    BEGIN OF ty_authorization_field,
```
GuilhermeSaraiva96, line 29: could you check with Yvonne if any of the fields in your AFF can be used from her shared interface?

file-formats/suso/type/zif_aff_suso_v1.intf.abap
```diff
@@ -0,0 +1,213 @@
...
+      "! <p class="shorttext">Authorization Field</p>
+      "! Name of the authorization field. An authorization object can contain up to 10 authorization fields.
+      "! $required
+      authorization_field TYPE c LENGTH 10,
+      "! <p class="shorttext">Description</p>
+      "! Description of authorization field
+      description         TYPE c LENGTH 60,
```
GuilhermeSaraiva96: I think this field does not belong to the SUCO AFF

file-formats/suso/type/zif_aff_suso_v1.intf.abap
```diff
@@ -0,0 +1,213 @@
...
+      not_classified TYPE ty_access_category VALUE ' ',
+    END OF co_access_category.
+
+  TYPES:
+    "! <p class="shorttext">Permitted Activities</p>
+    "! Permitted activities
+    BEGIN OF ty_permitted_activity,
+      "! <p class="shorttext">Activity</p>
+      "! Activity
+      activity        TYPE c LENGTH 2,
```
GuilhermeSaraiva96, line 85: would maybe a big enumeration for the activity fields (display,read,etc) make sense here? a string is also fine, but I think there is a fixed set of options.

file-formats/suso/type/zif_aff_suso_v1.intf.abap
```diff
@@ -0,0 +1,213 @@
...
+      "! <p class="shorttext">Description</p>
+      "! Description of the activity
+      description     TYPE c LENGTH 30,
```
GuilhermeSaraiva96, line 88: activity and its description are basically the same field, as they have a one-to-one relationship. I would choose just the activity code. and somehow for the UI through a value help or something

file-formats/suso/type/zif_aff_suso_v1.intf.abap
```diff
@@ -0,0 +1,211 @@
...
+    "! Authorization object
+    BEGIN OF ty_main,
+      "! <p class="shorttext">Format Version</p>
+      "! Format version
+      "! $required
+      format_version       TYPE zif_aff_types_v1=>ty_format_version,
+      "! <p class="shorttext">Header</p>
+      "! Header
+      "! $required
+      header               TYPE zif_aff_types_v1=>ty_header_60,
```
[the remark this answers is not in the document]
GuilhermeSaraiva96: good point, does it support keyUser?

file-formats/suso/type/zif_aff_suso_v1.intf.abap
```diff
@@ -0,0 +1,213 @@
...
+      "! For authorization objects with ABAP language version 'ABAP for Cloud Development', the authorization object
+      "! class will be set to 'CPAE' and cannot be changed.
+      "! Caution: Authorization objects that should not be used any more, regardless of which application area they are
+      "! from, should be assigned to object class 'AAAA' (obsolete authorization objects). This only applies to
+      "! authorization objects with ABAP language version 'Standard ABAP'.
+      "! $required
+      object_class             TYPE c LENGTH 4,
+      "! <p class="shorttext">Object Class Description</p>
+      "! Description of the authorization object class
+      object_class_description TYPE c LENGTH 60,
```
GuilhermeSaraiva96, line 23: the description most probably does not belong to this AFF, but rather to the object class

file-formats/suso/type/zif_aff_suso_v1.intf.abap
```diff
@@ -0,0 +1,213 @@
...
+    END OF ty_general_information.
+
+  TYPES:
+    "! <p class="shorttext">Authorization Fields</p>
+    "! Authorization fields
+    BEGIN OF ty_authorization_field,
+      "! <p class="shorttext">Authorization Field</p>
+      "! Name of the authorization field. An authorization object can contain up to 10 authorization fields.
```
GuilhermeSaraiva96, line 31: up to 10 authorisation fields? what does this have to do with the TYPE c LENGTH 10?

file-formats/suso/type/zif_aff_suso_v1.intf.json
```diff
@@ -0,0 +1,7 @@
+{
+  "formatVersion": "1",
+  "header": {
+    "description": "AFF type for SUSO - V1",
```
GuilhermeSaraiva96:
```suggestion
    "description": "AFF type for SUSO (Authorisation Object)",
```

file-formats/suso/type/zif_aff_suso_v1.intf.abap
```diff
@@ -0,0 +1,213 @@
...
+      "! Description of authorization field
+      description         TYPE c LENGTH 60,
+      "! <p class="shorttext">Activity Field</p>
+      "! By selecting this option, an authorization field can be flagged that can contain activity-related values.
+      "! The ACTVT field is always an activity field.
+      "! Organizational levels always represent organizational responsibilities, but never activities. Therefore, they
+      "! cannot be declared as activity fields.
+      "! If fields do not have a search help, it is very difficult to tell whether or not they can contain
+      "! activity-related values. Therefore, they cannot be declared as activity fields.
+      is_activity_field   TYPE abap_bool,
```
GuilhermeSaraiva96, line 41: I find it hard to understand if this field should rather belong to the AFF of the AUTH. It is information related to the field, not to SUSO, right?

Same for the search help

GuilhermeSaraiva96 (changes requested):

file-formats/suso/type/zif_aff_suso_v1.intf.abap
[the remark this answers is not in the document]
WDFdaniel: Yes, SUSO supports keyUser.

file-formats/suso/type/zif_aff_suso_v1.intf.abap
WDFdaniel: The mentioned restriction to a maximum of 10 authorization fields per authorization object has nothing to do with TYPE c LENGTH 10. However, this note could be helpful for the user since it can be displayed as tooltip in VS Code. Or would you suggest to remove, or rephrase this sentence?

GuilhermeSaraiva96:
> Notes: The structure and information in AFF are mainly based on the native SUSO ADT editor to cover the information available there. Object Class Description, Description of Authorization Fields (in ty_authorization_field) and Description of Activities (in ty_permitted_activity) will be read-only and be read from tables TOBCT, DD04T and TACTT. The Activity Field Flag (is_activity_field) and the Access Category (access_category) are only relevant in cloud systems and have connection to SIA3 object (Authorization Object Extension). For Activity Field Flag and Access Category BADI implementation from SIA3 called in context of the get and save processes in current SUSO ADT editor implementation. Would that be ok? Regarding the examples: The set of examples might be enhanced and renamed later on.

the AFF has nothing to do with the ADT Editors. The AFF represents a data model, which will be used afterwards by the SD-UI framework to draw an UI based on the skeleton of the data model, but the AFF is actually decoupled from the UI framework.

These descriptions, even though they are readonly, **cannot** coexist in any other AFF. If they belong to another AFF; they can't be here. The readonly property belongs to the UI, not to the AFF

file-formats/suso/type/zif_aff_suso_v1.intf.abap
WDFdaniel: I now reuse the type for authorization field from the shared interface. For other elements, I do not see a (useful) option to reuse something from a shared interface. There is also a type for the field values, but with a length of 40, but activities only have two characters.

file-formats/suso/type/zif_aff_suso_v1.intf.abap
WDFdaniel: I removed that field from AFF.

file-formats/suso/type/zif_aff_suso_v1.intf.abap
WDFdaniel: There is set of activities, but from time to time new activities are added and there are already more than 260 activities available. I think an enumeration or string would therefore not make much sense here.

file-formats/suso/type/zif_aff_suso_v1.intf.json
WDFdaniel: I replaced the '- V1' with '(Authorization Object)'.

file-formats/suso/type/zif_aff_suso_v1.intf.abap
WDFdaniel: is_activity_field refers to the field, but does not belong to it. is_activity_field as well as access_category are part of the SIA3 (Authorization object extension) integration that has been started but not yet completed.

Regarding the Search Help:
The Search Help Field belongs to SUSO (stored in table TOBJ_FLD_EXT). It’s an Object-Field Search Help, so each object can define a search help if needed. In AUTH AFF this information is actually also used but as a Boolean value, we do not show the name (in principle, different SUSO objects can define different search helps for the same AUTH Field).

file-formats/suso/type/zif_aff_suso_v1.intf.abap
WDFdaniel: Yes, however, there is no AFF (and also no editor) for the object class (it can only be maintained via SU21 in the backend - so in the cloud not at all), and I don't know of any plan to provide one for it either.

file-formats/suso/type/zif_aff_suso_v1.intf.abap
WDFdaniel: When adding an activity to the list, a value help is definitely useful, and it is also intended here. But when someone only displays the list, activity codes (at least the ones that are not used so frequently) might be incomprehensible for the user without the description. So from our perspective it makes sense to display the description here, so it becomes clear to the user what the meaning of the codes is without the need to look them up somehow one by one.

file-formats/suso/type/zif_aff_suso_v1.intf.abap
```diff
@@ -0,0 +1,210 @@
...
+    END OF ty_general_information.
+
+  TYPES:
+    "! <p class="shorttext">Authorization Fields</p>
+    "! Authorization fields
+    BEGIN OF ty_authorization_field,
+      "! <p class="shorttext">Authorization Field</p>
+      "! Name of the authorization field. An authorization object can contain up to 10 authorization fields.
+      "! $required
+      authorization_field TYPE if_aff_auth_types_v1=>ty_auth_field,
```
GuilhermeSaraiva96: what is this if_aff_auth_types_v1?

GuilhermeSaraiva96 (changes requested):

file-formats/suso/type/zif_aff_suso_v1.intf.abap
WDFdaniel: This is an shared interface (it is also used in AFF for SUSH), which I reuse here.

GuilhermeSaraiva96: it does not exist, it is missing the Z....

WDFdaniel: Thanks! I changed the name.

file-formats/suso/type/zif_aff_suso_v1.intf.abap
GuilhermeSaraiva96: I understand, but in my perspective it does not matter as it does not belong to this Aff. If there would ever be an aff for that object class, then the Suso aff would be invalid. Imagine if we did this for other object types that turned out to be implemented in the future. Tell me if there are any other fields in the same situation, this is a critical problem in the AFF

file-formats/suso/type/zif_aff_suso_v1.intf.abap
GuilhermeSaraiva96: I do not see this change being based off the SUSH change

file-formats/suso/type/zif_aff_suso_v1.intf.abap
GuilhermeSaraiva96: ah the 10 is a coincidence. Then it is fine, thought it would be an error in the tooltip

@albertmink there is a json property called "maxItems". This limits the number of items in an array. What do you think of adding support for this, instead of having a tooltip?

file-formats/suso/type/zif_aff_suso_v1.intf.abap
GuilhermeSaraiva96: how would the AI then know which activities are supported? we could keep this enumeration up to date, by having a script syncing with the backend. Or if there is a place in the backend with this information, the AI can also have access to it

The goal is for the AI to create and maintain SUSO's in the future.

file-formats/suso/type/zif_aff_suso_v1.intf.abap
GuilhermeSaraiva96: The AFF is a data model, you are still thinking of the AFF as a pure UI tool. The AFF is used to transport data between systems and to represent an instance of the object type. The description brings no advantage whatsoever in these cases, it is only useful to show extra information to the user. This is a clear sign that the field does not belong to the AFF. What happens if the user updates the activity, but not the description field? The user can even be the AI, or someone using the source editor. In these cases, you cannot have a side effect and  force both to be in sync.

Please focus just on the AFF, the UI is something we can solve afterwards. Again please check if any other field in the AFF is the same conditions as this one. I cannot be flagging all these fields, when I do not know the object type

GuilhermeSaraiva96: We can maybe add element info or something, there are many ways. what we cannot have is redundant info in an AFF

file-formats/suso/type/zif_aff_suso_v1.intf.abap
WDFdaniel: Within this type definition ty_authorization_field, I use a type from shared interface for typing authorization_field. This shared interface is part of the SUSH PR.

file-formats/suso/type/zif_aff_suso_v1.intf.abap
WDFdaniel: The activities codes that are relevant at this point are all stored in table TACT. Would this be sufficient for the AI or would we need something else?
Permitted activities shall only be maintained if we have authorization field ACTVT in the field list. If this field is not part of field list, no permitted activities must be maintained.

file-formats/suso/type/zif_aff_suso_v1.intf.abap
WDFdaniel: Ok, I remove the description.

file-formats/suso/type/zif_aff_suso_v1.intf.abap
WDFdaniel: Ok, I remove that description.

file-formats/suso/type/zif_aff_suso_v1.intf.abap
WDFdaniel: Regarding other fields (you asked in a comment above): is_activity_flag and also access_category is related to started SIA3 integration. The content is stored in table(s) that not belong to SUSO data model, however SUSO is connected via BAdI implementation. Would that be also a problem?

WDFdaniel:
I removed the SIA3-integration related fields (is_activity_field and access_category) from AFF. Now it only contains fields that belong to SUSO.
Checks are still failing, but I reused a type from an interface which is part of the SUSH PR.

WDFdaniel:
Added the two fields is_activity_field and access_category back. A migration from SIA3 to SUSO is planned (the fields are assumed to remain and SIA3 will be removed after migration) and there will be no extra integration for SIA3 (there is the integration in the existing native ADT editor for SUSO, where they can be read and maintained), but they need to be maintainable.

file-formats/suso/type/zif_aff_suso_v1.intf.abap
GuilhermeSaraiva96: @WDFdaniel there is now a new annotation called maxItems, und minItems. Here is the documentation: https://github.com/SAP/abap-file-formats/blob/main/docs/json.md#number-of-items

GuilhermeSaraiva96 (changes requested):

file-formats/suso/type/zif_aff_suso_v1.intf.abap
```diff
@@ -0,0 +1,205 @@
...
+      "! from, should be assigned to object class 'AAAA' (obsolete authorization objects). This only applies to
+      "! authorization objects with ABAP language version 'Standard ABAP'.
+      "! $required
+      object_class TYPE c LENGTH 4,
+    END OF ty_general_information.
+
+  TYPES:
+    "! <p class="shorttext">Authorization Fields</p>
+    "! Authorization fields
+    BEGIN OF ty_authorization_field,
```
GuilhermeSaraiva96: missing the fields from SIA3, right?

file-formats/suso/examples/z_aff_example_suso_1.suso.json
```diff
@@ -0,0 +1,39 @@
...
+  "header": {
+    "description": "Access to Configuration Objects",
+    "originalLanguage": "de"
+  },
+  "generalInformation": {
+    "objectClass": "BC_A"
+  },
+  "authorizationFields": [
+    {
+      "authorizationField": "ACTVT"
```
GuilhermeSaraiva96: isActivityField = true?

file-formats/suso/type/zif_aff_suso_v1.intf.abap
```diff
@@ -0,0 +1,205 @@
...
+      "! $required
+      authorization_field TYPE zif_aff_auth_types_v1=>ty_auth_field,
+      "! <p class="shorttext">Activity Field</p>
+      "! By selecting this option, an authorization field can be flagged that can contain activity-related values.
+      "! The ACTVT field is always an activity field.
+      "! Organizational levels always represent organizational responsibilities, but never activities. Therefore, they
+      "! cannot be declared as activity fields.
+      "! If fields do not have a search help, it is very difficult to tell whether or not they can contain
+      "! activity-related values. Therefore, they cannot be declared as activity fields.
+      is_activity_field   TYPE abap_bool,
```
GuilhermeSaraiva96: Can more than 1 field be set as isActivityField=true?

file-formats/suso/examples/z_aff_example_suso_1.suso.json
wurzka: This is a duplication of `authorizationField`.
I suggest 
```suggestion
  "authorizationFields": [
    {
      "name": "ACTVT"
```
or 

```suggestion
  "authorizationFields": [
    {
      "fieldName": "ACTVT"
```

file-formats/suso/type/zif_aff_suso_v1.intf.abap
```diff
@@ -0,0 +1,205 @@
...
+      others         TYPE ty_access_category VALUE 'O',
+      "! <p class="shorttext">Not Classified</p>
+      "! Not classified
+      not_classified TYPE ty_access_category VALUE ' ',
+    END OF co_access_category.
+
+  TYPES:
+    "! <p class="shorttext">Permitted Activities</p>
+    "! Permitted activities
+    BEGIN OF ty_permitted_activity,
```
GuilhermeSaraiva96: This is only valid if there is a field which is activity, right?

GuilhermeSaraiva96 (changes requested):
still to clarify if the sia3 fields belong to the aff. the answer is not as easy as I thought

PR #857, opened 2026-09-08
Fix broken editorconfig-checker action (upgrade to v3)

GuilhermeSaraiva96:
## Summary
- Upgrades `editorconfig-checker/action-editorconfig-checker` from `@v1` to `@v2`

## Reason
The v1 Docker image is broken — the `ec` binary is missing, causing every PR's editorconfig check to fail with:
```
/entrypoint.sh: line 3: ec: not found
```
This is an upstream issue with the action, not a problem in any PR's files. Upgrading to v2 resolves it.

### DIFF
--- a/.github/workflows/editorconfig.yml
+++ b/.github/workflows/editorconfig.yml
@@ -32,4 +32,4 @@ jobs:
       - name: Checkout the repository
         uses: actions/checkout@v6
       - name: Check for editorconfig violations
-        uses: editorconfig-checker/action-editorconfig-checker@v1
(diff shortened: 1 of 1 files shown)


GuilhermeSaraiva96:
the problem can be seen in recently built PRs, error: /entrypoint.sh: line 3: ec: not found

albertmink (approved):
Thanks @GuilhermeSaraiva96  let's move version.

albertmink:
LGTM on https://github.com/SAP/abap-file-formats/actions/runs/34200281886/job/101977255834?pr=854 👌

PR #863, opened 2026-09-16
[CHKV] Make selectedChecks required

Markus1812:
A check variant (CHKV) without any checks cannot exist and raises syntax errors from the backend.

Compatibility concern: As a check variant with no checks cannot be activated, there should not be any CHKV JSON where this property is missing.


GuilhermeSaraiva96:
> I didn’t post inline comments. The main remaining concern is that making `selectedChecks` required still allows `"selectedChecks": []`, so if the backend rejects check variants without checks, the schema may still permit an invalid CHKV unless the format has a way to express a non-empty array constraint.
> 
> PR Bot Information

I was also wondering about this. How do we enforce a minimum amount of checks of 1?

Markus1812:
> I was also wondering about this. How do we enforce a minimum amount of checks of 1?

The backend syntax check is already doing this. 

The JSON Schema has a property called `minItems` (and `maxItems`) but the AFF currently does not support it. Reference: https://json-schema.org/understanding-json-schema/reference/array#length

GuilhermeSaraiva96:
One of the AUTH objects also needs the maxItems annotation. I think it would be a good idea to implement this, should be low effort  @schneidermic0

GuilhermeSaraiva96 (dismissed):

GuilhermeSaraiva96 (approved):

PR #864, opened 2026-09-21
generation: dont pull, just do git clone

larshp:
https://github.com/SAP/abap-file-formats-tools/pull/390#issuecomment-5758226938


Markus1812 (approved):
Thanks

PR #866, opened 2026-09-23
feat(chko): mark parameter name and hidden as required

albertmink:
## Problem

The `hidden` flag on a CHKO parameter is not serialized when `false`, so it disappears from the JSON representation and only reappears after the ADT "Import Parameters" action. `name` is the unique key of `ty_parameters` but was likewise not enforced.

## Change

Add `$required` to `name` and `hidden` in `ty_parameter`. This makes both fields always serialized (independent of their value) and adds them to the generated schema's `required` array.

- `type/zif_aff_chko_v1.intf.abap`: `$required` annotations
- `chko-v1.json`: regenerated `required: [name, hidden]`
- example: added `hidden: false` to stay schema-valid

🤖 Generated with Claude Code


albertmink:
Superseded by #867 (branch renamed to feat/ prefix to avoid leaking an internal ID).

